PR #13, opened 2011-08-08
New "generate" command

weaverryan:
Hey guys!

This is a basic generate command for generating a new fixture class. It's basic and has really just one option: whether or not it should be container aware.

Also, small updates for the other task.

Thanks!


lavoiesl:
Wow, it is indeed very old… Moving it to `SensioGeneratorBundle` would be nice indeed.

fabpot:
As creating fixtures has been simplified in #189, I'm not sure this is still relevant. Closing.

PR #40, opened 2012-04-08
I was created a new option, "bundles"

fchris82:


fabpot:
I'm closing this old PR as Symfony Flex will remove the need for bundles, so I don't think this should be needed anymore.

PR #67, opened 2012-11-17
Allow fixture classes to be loaded individually.

mariusz-alef-bak:
This pull requests allows to load individual fixture classes by their class names by adding `--fixture-classes` option.

In some projects which require various fixture sets (e.g. for different setups for various environments) being able to point fixtures only by directory is quite a pain as some fixtures may be reused in few sets. Moving them to subdirectories doesn't solve the problem in some cases as namespaces would have to be adjusted accordingly and code would have to be duplicated. There were some issues reported concerning the lack of this feature both in this repo and in data-fixtures.

I have made a similar pull request https://github.com/doctrine/data-fixtures/pull/79 to data-fixtures  to allow loading individual files instead of directories and I was hoping to add this behaviour also to DoctrineFixturesBundle, but the pull request hasn't been merged yet. This pull request is another way to add similar (or maybe even better) functionality directly to DoctrineFixturesBundle, as `addFixture` method is already available in data-fixtures `Loader` class.

### DIFF
--- a/Command/LoadDataFixturesDoctrineCommand.php
+++ b/Command/LoadDataFixturesDoctrineCommand.php
@@ -40,7 +40,8 @@ class LoadDataFixturesDoctrineCommand extends DoctrineCommand
         $this
             ->setName('doctrine:fixtures:load')
             ->setDescription('Load data fixtures to your database.')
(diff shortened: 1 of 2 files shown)


lavoiesl:
Duplicate of #68

PR #86, opened 2013-05-31
Add documentation for DependentFixtureInterface

jbafford:
Resources/doc/index.rst gives an example using the OrderedFixtureInterface; the DependentFixtureInterface should also be described.

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -280,6 +280,53 @@ by ``getOrder()``. Any object that is set with the ``setReference()`` method
 can be accessed via ``getReference()`` in fixture classes that have a higher
 order.
 
+You can also specify explicit dependencies with the ``DependentFixtureInterface``:
+
+.. code-block:: php
+
+    // src/Acme/HelloBundle/DataFixtures/ORM/LoadUserGroupData.php
+
+    namespace Acme\HelloBundle\DataFixtures\ORM;
+
+    use Doctrine\Common\DataFixtures\AbstractFixture;
+    use Doctrine\Common\DataFixtures\DependentFixtureInterface;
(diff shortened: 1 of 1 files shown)


lavoiesl:
Isn’t the documentation from the library enough? https://github.com/doctrine/data-fixtures#dependentfixtureinterface

jbafford:
Apparently, I didn't think so at the time.

Especially since both of these packages are part of the Doctrine project, I think that if this bundle is going to provide documentation regarding how to use the data-fixtures package, it should be complete. Alternatively, don't provide any documentation at all on how to use data-fixtures, and instead, clearly point to the data-fixtures package documentation. (There are links to the data-fixtures package, but none that clearly say, Read The Documentation There.)

Otherwise, with a partial documentation set, you run the risk of confusing people who see this bundle as the entry point for the functionality, and don't think to look at the documentation for the main package.

At the time, I must have figured something along the lines of, "well, this is part of Doctrine, so clearly they're going to have all the relevant documentation here", and saw the lack of the DependentFixtureInterface documentation as an oversight. If I had seen the data-fixture's documentation at the time, this commit might instead just be along the lines of, "you can do this too, see the DependentFixtureInterface documentation for details". If you think that would be a better addition to the docs than this one, I can make a new PR for that.

lavoiesl:
I agree with you that there should not be partial documentation. However, I believe it would be best to put the general documentation inside the general library and simply point to it. Complete examples of using DoctrineFixturesBundle could also become a Cookbook entry.

stof:
I agree with @lavoiesl here. The documentation about features provided by the library should go in the library.
The bundle is only about exposing these features in the console of Symfony projects.

guilhermeblanco:
This type of documentation should go to the library, closing.

PR #105, opened 2013-10-15
New command added to get the last order used in OrderedFixtures

mangelsnc:
# Get Last Order

This command let known what is the last index used in _OrderedFixtures_. 

When we are working in team in a huge project, often we use repeated indexes for the fixtures we create because don't remember which is the last used index. 

With this command, now we are able to know it. It's specially useful in projects with a lot of fixtures.

### DIFF
--- a/Command/GetLastOrderCommand.php
+++ b/Command/GetLastOrderCommand.php
@@ -1,0 +1,72 @@
+<?php
+
+/*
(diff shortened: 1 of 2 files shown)


fabpot:
As there is the possibility to use `DependentFixtureInterface` instead of integer orders, I think this is not needed anymore. So, I'm closing this PR.

PR #106, opened 2013-10-15
List fixtures

mangelsnc:
# List Fixtures

Added a new command which lists all the registered fixtures.

If the fixture implements OrderedFixture it shows it's order otherwise shows a dash


fabpot:
Closing as #189 made huge improvements in usability and as we are now using `DependentFixtureInterface`, there is no need to know the order anymore.

PR #111, opened 2013-12-12
Highlight the default value in the confirmation message

rybakit:


guilhermeblanco:
fixed on a previous commit

PR #113, opened 2014-01-29
This provides an option to use one transaction per fixture file

gnat42:
In tandem with my pull request on doctrine/data-fixtures this allows a user to select whether they want to load fixtures in a single transaction or one transaction per fixture


fabpot:
reverted

OskarStark:
why did you remove this without deprecating it?
Now my command fails oO

fabpot:
That did not work, it was never implemented.

PR #117, opened 2014-04-04
add / foreign key checks switch

scasei:
Because while purging the database there is a foreign key check error often


guilhermeblanco:
Can't merge this as-is, since this is MySQL specific and FixturesBundle works with more DBs than only MySQL.
Please rework on this to make it generic and re-submit for our consideration.

Thanks

PR #120, opened 2014-06-04
add a getter method for retrieving new instances of ORMPurger

briancappello:
The motivation behind this PR is to allow for overriding the doctrine:fixtures:load command and using a custom ORMPurger class. It does not cause any BC breaks.

See https://github.com/kayue/KayueWordpressBundle/pull/37 for an example of a real-world use case.

### DIFF
--- a/Command/LoadDataFixturesDoctrineCommand.php
+++ b/Command/LoadDataFixturesDoctrineCommand.php
@@ -97,12 +97,23 @@ EOT
                 sprintf('Could not find any fixtures to load in: %s', "\n\n- ".implode("\n- ", $paths))
             );
         }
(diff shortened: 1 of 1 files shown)


guilhermeblanco:
What's the purpose of doing this?

briancappello:
@guilhermeblanco Feel free to close this PR; the original intention was to allow creating a custom ORMPurger that wouldn't delete *all* tables in the database, however, we were able to solve this by not using doctrine's auto-mapping feature and creating multiple entity managers instead, as per stof's suggestion.

guilhermeblanco:
Ok... thanks! =)

PR #131, opened 2015-01-02
Update composer.json

chadyred:
Hi,

the dependencie of this bundle on his last version "doctrine/doctrine-fixtures-bundle": "2.2.*@dev" with "doctrine-bridge" is out of date : "symfony/doctrine-bridge": "~2.1" ("2013-05-26 18:42 UTC").

Can I inject the dependencie with the last version of doctrine-bridge which is "symfony/doctrine-bridge": "3.0.*@dev" (2014-12-30 10:17 UTC) ?

Thanks stof !

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     ],
     "require": {
         "php": ">=5.3.2",
-        "symfony/doctrine-bridge": "~2.1",
+        "symfony/doctrine-bridge": "~3.0",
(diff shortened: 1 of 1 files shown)


stof:
this is wrong. Symfony 3.0 is far from being released (it will be only released at the end of November 2015), and there will still be new released of 2.x before that (2.7 is coming in May).

PR #134, opened 2015-02-22
Allow symfony 3.0 components

nicolas-grekas:
Tests should tell if any deprecated interfaces of Symfony are used. If not, then the bundle is defacto compatible with 3.0


sroze:
@nicolas-grekas and it's the case, see #138

PR #138, opened 2015-04-09
Remove usage of deprecated `dialog` helper when `question` is available

sroze:
This PR remove usage of command "Dialog" helper and use the "Question" helper if available.

/cc @stof


stof:
Thank you @sroze.

PR #142, opened 2015-06-13
Add help to avoid ContextErrorException

zatikbalazs:
| Q             | A
| ------------- | ---
| Doc fix?      | yes
| New docs?     | no
| Applies to    | all
| Fixed tickets |

It's a common problem to get a ContextErrorException when loading
fixtures with references. Adding the __toString() method to all
corresponding entity classes solves this problem.

I added a short note on how to avoid the ContextErrorException.

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -286,6 +286,24 @@ by ``getOrder()``. Any object that is set with the ``setReference()`` method
 can be accessed via ``getReference()`` in fixture classes that have a higher
 order.
 
+.. note::
+
+    If you get a ``ContextErrorException`` when trying to load the fixtures, then
+    add the ``__toString()`` method to all of the corresponding entity classes:
+
+    .. code-block:: php
+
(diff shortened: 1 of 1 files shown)


stof:
Why is that ? Entities don't need to be castable as strings

zatikbalazs:
You should try it with and without the ``__toString()`` method as well. For me it only worked with ``__toString()``.

guilhermeblanco:
One thing does not correlate to the other. `ContextErrorException` is caused by Symfony Debug component and it's up to that component address the issue, not this one.

zatikbalazs:
Alright then. I just wanted to add a note in the documentation since I followed the doc of DoctrineFixturesBundle and this exception came up every time. Even though it has something to do with the Symfony Debug Component, I thought it wouldn't hurt to mention this issue (with a solution) in the documentation of DoctrineFixturesBundle so that others don't have to waste hours to fix this issue like I did. It would have been a nice gesture and useful for other developers.

PR #147, opened 2015-09-13
using array_map in favor of foreach loop

Ma27:


guilhermeblanco:
This is not addressing anything in particular. Also the performance improvement is negligible considering this command is not used intensely on production.

PR #149, opened 2015-10-16
Documentation: composer require with --dev

fain182:


KingCrunch:
But maybe a staging system, that should run with production settings, needs them?

fain182:
Well, i wouldn't say it's a common case, usually staging use a dump of production, not some fixtures.

And anyway already when is inserted in the `app/AppKernel.php` is positioned inside the if of the development or test environment (line 45)

guilhermeblanco:
`--dev` is default now AFAIK, closing

PR #152, opened 2015-12-05
Moved loading of fixtures to a service

### LINKED ISSUE #88: Making a service out of doctrine:fixtures:load command
Would be great to have the command as service or to extract pure service class out of it in order to have it available in the container

jongotlin:
# symfonycon Hackday pr

Fixes #88 

Notice, I've removed `multiple-transactions` since it doesn't look like its working. Added in #113. Please correct me if I'm wrong.


alcaeus:
Handled in #209.

PR #155, opened 2015-12-19
Use ContainerAwareTrait instead of own implementation

### LINKED ISSUE #154: Use ContainerAwareTrait instead of own implementation
The new ``ContainerAwareTrait`` should be used instead of an own implementation.

Example:
https://symfony.com/doc/master/bundles/DoctrineFixturesBundle/index.html#using-the-container-in-the-fixtures

This:
```php
class LoadUserData implements FixtureInterface, ContainerAwareInterface
{
    /**
     * @var ContainerInterface
     */
    private $container;

    public function setContainer(ContainerInterface $container = null)
    {
        $this->container = $container;
    }
```

Should be changed to:
```php
class LoadUserData implements FixtureInterface, ContainerAwareInterface
{
  use ContainerAwareTrait;
```

JHGitty:
Fix #154

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -284,20 +284,12 @@ component when checking it:
     use Doctrine\Common\DataFixtures\FixtureInterface;
     use Doctrine\Common\Persistence\ObjectManager;
     use Symfony\Component\DependencyInjection\ContainerAwareInterface;
-    use Symfony\Component\DependencyInjection\ContainerInterface;
+    use Symfony\Component\DependencyInjection\ContainerAwareTrait;
     use AppBundle\Entity\User;
 
     class LoadUserData implements FixtureInterface, ContainerAwareInterface
     {
-        /**
-         * @var ContainerInterface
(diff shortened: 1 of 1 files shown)


xelaris:
I think the current code example should be kept, as the bundle is compatible with SF 2.3 (the trait was added in 2.4) and supports PHP 5.3.2. There is a note about not implementing the interface but extending ``Symfony\Component\DependencyInjection\ContainerAware`` at the bottom already. Maybe the option to use the trait should be mentioned there, along with mentioning the required versions of SF 2.4 and PHP 5.4. What do you think?

hobbysoldat:
Well, now we have a problem, because the Bundle is not compatible with Symfony 3 anymore (container is not injected). Maybe we could try to use duck typing to check if there is a setContainer Method in the Class and inject it if that is the case?

JHGitty:
Won't fix: https://github.com/doctrine/DoctrineFixturesBundle/issues/154

PR #160, opened 2016-01-28
Added environment to avoid unexpected result.

cn007b:
I read all documentation but i hack my dev db when tweak this bundle.)
That's why i decided that it will convenient to specify environment explicitly to emphasize one more time that fixtures will be inserted to test db.

Hope it will be helpful.

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -109,13 +109,13 @@ line by using the ``doctrine:fixtures:load`` command:
 
 .. code-block:: bash
 
-    php app/console doctrine:fixtures:load
+    php app/console doctrine:fixtures:load -env=test
 
 If you're using the ODM, use the ``doctrine:mongodb:fixtures:load`` command instead:
 
 .. code-block:: bash
 
-    php app/console doctrine:mongodb:fixtures:load
+    php app/console doctrine:mongodb:fixtures:load -env=test
 
 The task will look inside the ``DataFixtures/ORM/`` (or ``DataFixtures/MongoDB/``
 for the ODM) directory of each bundle and execute each class that implements


fabpot:
Fixtures are not just for testing. They can also be used to bootstrap dev envs, or even production when deploying the first time.

cn007b:
Yes, agree with you! But, I've found [this](http://prntscr.com/gkafc9) on page: https://github.com/doctrine/DoctrineFixturesBundle/blob/dd52aa9230ba5da5fe02f61fda5833df7bf136b9/Resources/doc/index.rst
How come?

Resources/doc/index.rst
```diff
@@ -109,13 +109,13 @@ line by using the ``doctrine:fixtures:load`` command:
 
 .. code-block:: bash
 
-    php app/console doctrine:fixtures:load
+    php app/console doctrine:fixtures:load -env=test
```
stof: this won't work. It must be `-e` or `--env`

stof:
well, the page you linked is seeing the doc for **your** commit.

The fact that it is accessible in the doctrine repository even though it is not reachable by any branch there is due to the fact that all forks are stored as a single git repo in Github, with a special handling of references (emulating the fact that each repo has its own branches).
On a commit page, Github displays in which branches the commit is reachable (after the commit message but before the author, see https://github.com/doctrine/DoctrineFixturesBundle/commit/7bb198c044b798b54e6be37c7929339aa645c3bf). The link you gave does not give any branches allowing to reach your commit.

cn007b:
My fault...
I've deleted my message. Thank you! And sorry for this stupid situation...

PR #166, opened 2016-04-22
Update index.rst

pyatnitsev:
in Symfony 3.0 Console was moved to bin directory, but in docs it still in app/

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -109,13 +109,13 @@ line by using the ``doctrine:fixtures:load`` command:
 
 .. code-block:: bash
 
-    php app/console doctrine:fixtures:load
+    php bin/console doctrine:fixtures:load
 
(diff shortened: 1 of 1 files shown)


stof:
@javiereguiluz what is the policy for bundle docs on symfony.com ? should we use bin/console or app/console in the doc ? The app/console location is not dead yet (and won't be for years)

javiereguiluz:
The problem is that we don't have a policy ... and this seemingly insignificant change is causing LOTS of troubles for newcomers. Some bundles are showing both alternatives: https://github.com/javiereguiluz/EasyAdminBundle#step-4-prepare-the-web-assets-of-the-bundle ... but that's ugly 😞

pyatnitsev:
Maybe better solution is write 2 variants of this commands, for symfony 3 and for symfony 2.

Something like this:
If you use symfony2....
And if you use symfony 3...

fabpot:
already merged as #194

PR #167, opened 2016-04-29
Update AppKernel installation doc

JHGitty:
see https://github.com/symfony/symfony-standard/blob/master/app/AppKernel.php#L21


fabpot:
already merged by another PR.

PR #169, opened 2016-06-01
Fixtures list command

miholeus:
Sometimes when having many fixtures you end up with lots of files and don't know which one is executed first and the second. We use `OrderedFixtureInterface` to set order explicitly, but what will you do if you need to change order of some fixture? You'll need in looking up for order in many files, that's tedious work. These PR solves that problem, you'll always see list of available migrations ordered. Also you can find migration by selected order.
So new command is added to bundle:
`doctrine:fixtures:list`

### DIFF
--- a/Command/AbstractFixturesCommand.php
+++ b/Command/AbstractFixturesCommand.php
@@ -1,0 +1,87 @@
+<?php
+
+/*
+ * This file is part of the Doctrine Fixtures Bundle
+ *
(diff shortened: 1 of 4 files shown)


miholeus:
Will somebody merge it?

fabpot:
#189 made this obsolete I think as we are implementing `DependentFixtureInterface`. So, integer orders should not be used anymore.

PR #175, opened 2016-08-03
allow overriding of services in LoadDataFixturesDoctrineCommand (branch master)

fprochazka:
I wanted to override something in executor, but that cannot be done without copy&pasting huge chunks of code in the command. This simple refactoring solves it.


Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -139,4 +141,32 @@ private function askConfirmation(InputInterface $input, OutputInterface $output,
 
         return $questionHelper->ask($input, $output, $question);
     }
+
+	/**
```
TomasVotruba: It looks like coding standard is miss matched.

alcaeus:
I'm closing this PR: for one, we don't want to make the purgers and executors injectable this way. As I've explained in https://github.com/doctrine/DoctrineFixturesBundle/issues/88#issuecomment-499841615, this needs changes in the underlying library, after which we'll be able to create services for purgers and executors which people can then override based on interfaces. In the meantime, I'm closing this PR. Thanks @fprochazka, and sorry for the very, very, very late feedback.

PR #178, opened 2016-09-12
Add a missing role for admin

mik-laj:


fabpot:
Those are just examples. Anyway, as we moved away from user to products in the docs, that's not relevant anymore.

Issue #177, opened 2016-08-05
Running doctrine:fixtures:load truncates migration_versions table


frak:
When we instantiate a project for a new developer, we first run app/console doctrine:migrations:migrate to load the schema. We then run doctrine:fixtures:load to generate the test data for development. However, after that the migrations table has been truncated, so if a developer then creates a migration as part of their feature, the only way we can run that migration is to drop the database completely and reload the fixtures.

I am struggling to find a use case for this behaviour, and am wondering if there is any way to get the fixtures command to skip the migrations table?

damlys:
I have pushed the feature that can help you
https://github.com/doctrine/DoctrineFixturesBundle/pull/185/commits

alcaeus:
Looking at the ORM purger, it only truncates table based on mapped entities. It will only clear the migrations table if you've mapped that as an entity in ORM: https://github.com/doctrine/data-fixtures/blob/03ee528bfbb1b315ee8fb387ba95ad5a84d27078/lib/Doctrine/Common/DataFixtures/Purger/ORMPurger.php#L114..L118. Since the table is internal, we don't recommend mapping it as an entity: you shouldn't need to read from it and you most definitely shouldn't just write data to it.

Closing as `Wontfix`.

yahyaerturan:
Hi,

I know this an old issue but I have similar problem:

In my project, I have Country, City, County entities. If I feed them with real data through fixtures, it re-writes each time - which is unnecessary. If I feed them with migrations, fixtures.load reset those nicely.

What is the solution for that? Is it really a bad practice to have an option tables_to_be_not_purged option?

PR #185, opened 2017-01-03
Configuration allows exclude tables from purge

damlys:


alcaeus:
Not a fan of using an option for this - if you really want to exclude tables from being purged, it should probably be added as a command option. Given that the original issue (#177) based this on an invalid use-case, I'm closing this pull request.

PR #192, opened 2017-04-26
Add kernel root directory to fixtures location paths

baleyko:
It would be helpful to handle fixtures automatically(without specifying of path), in applications those are using MicroKernelTrait or similar ones, those do not define bundles in itself.
Small fix great profits.


javiereguiluz:
This one is very important for Symfony Flex and Symfony 4. This issue bugged us when upgrading the Symfony Demo to Flex (see https://github.com/symfony/symfony-demo/pull/617).

@baleyko thanks for contributing this fix!

PR #201, opened 2017-09-14
Remove `--dev` option of installation

issei-m:
Hi, I’ve been started using Symfony 3.3, and DIC auto configurator, and I just realized DoctrineFixturesBundle document is probably obsolete.

The DIC auto configurator collects classes located in specified dir and tries to make the reflections of them.
However it must be failed with `ReflectionException` during reflecting classes in each bundles’ DataFixtures dir because `doctrine/fixture` or `doctrine/doctrine-fixtures-bundle` may not be installed on non-dev env (According to below docuement, they’d be only installed on dev).

http://symfony.com/doc/current/bundles/DoctrineFixturesBundle/index.html

Thanks.

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -17,7 +17,7 @@ following command to download the latest stable version of this bundle:
 
 .. code-block:: bash
 
-    composer require --dev doctrine/doctrine-fixtures-bundle
+    composer require doctrine/doctrine-fixtures-bundle
 
 This command assumes you have Composer installed globally, as explained
 in the `installation chapter`_ of the Composer documentation.


issei-m:
In addition to, Symfony's Best practices doc isn't using --dev flag anyway.

issei-m:
@fabpot Execuse me, could you review this? Actually, I'm not sure who lead maintainer of this bundle is though, the almost latest commits have been made by you, so I decided to choose you. Thanks.

alcaeus:
TBH, I believe the bundle should be used in development only. The docs correctly reflect this by requiring this bundle as a dev-dependency and only enabling it in the dev environment. Closing as `Wontfix`

PR #205, opened 2017-10-21
Fix sf3.4 command autoregistration deprecation

### LINKED ISSUE #204: Command auto-registration is deprecated since Symfony 3.4
Spotted while running tests in an app upgraded to Symfony 3.4:

```
Auto-registration of the command
"Doctrine\Bundle\FixturesBundle\Command\LoadDataFixturesDoctrineCommand" is
deprecated since Symfony 3.4 and won't be supported in 4.0. Use PSR-4 based
service discovery instead.
```

ogizanagi:
Fixes #204

> Auto-registration of the command "Doctrine\Bundle\FixturesBundle\Command\LoadDataFixturesDoctrineCommand" is deprecated since Symfony 3.4 and won't be supported in 4.0. Use PSR-4 based service discovery instead

### DIFF
--- a/Resources/config/services.xml
+++ b/Resources/config/services.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" ?>
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+    <services>
+
+        <!-- commands -->
+        <service id="doctrine.fixtures_load_command" class="Doctrine\Bundle\FixturesBundle\Command\LoadDataFixturesDoctrineCommand">
+            <tag name="console.command" command="doctrine:fixtures:load" />
+        </service>
+
+    </services>
+</container>
--- a/DependencyInjection/DoctrineFixturesExtension.php
+++ b/DependencyInjection/DoctrineFixturesExtension.php
@@ -1,0 +1,33 @@
+<?php
+
+/*
+ * This file is part of the Doctrine Fixtures Bundle
+ *
+ * The code was originally distributed inside the Symfony framework.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ * (c) Doctrine Project
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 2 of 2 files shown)


DoctrineFixturesBundle.php
```diff
@@ -24,4 +25,11 @@
  */
 class DoctrineFixturesBundle extends Bundle
 {
+    /**
+     * {@inheritDoc}
+     */
+    public function registerCommands(Application $application)
+    {
+        // noop
```
javiereguiluz: As I said in the related pull request made by @chalasr in Symfony repo, this may be technically correct but it looks super strange. A method to register commands that must be completely empty to register the commands?

ogizanagi: You're right of course. Removed.

Resources/config/services.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" ?>
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+    <services>
+
+        <!-- commands -->
+        <service id="doctrine.fixtures_load_command" class="Doctrine\Bundle\FixturesBundle\Command\LoadDataFixturesDoctrineCommand">
```
chalasr: should be private

ogizanagi: Actually no, as we need to to support older versions. Hence getting the command from the container directly rather than from the command loader.

ogizanagi: Hmm, NVM. I forgot a public alias was created in such case, so should be okay.

ogizanagi: Reverted. See:

https://github.com/symfony/symfony/blob/2661da415e108a479589bb943d4c12c8088c6e38/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/AddConsoleCommandPass.php#L31-L33

I knew [I already hit](https://travis-ci.org/lexik/LexikJWTAuthenticationBundle/jobs/290768849#L555) the issue somewhere 😉

shreypuranik:
I had a similar issue, updating doctrine/doctrine-bundle resolved it. Many thanks for merging this fix in @fabpot.

PR #209, opened 2017-11-24
[3.0] Re-working fixtures loading to load tagged services

weaverryan:
Hi guys!

This is very simple and is important because it will be very annoying to use this bundle in 4.0 when all services are private. Actually, in 3.3, it is already annoying :).

A) Now, fixtures are not loaded from directories. They are loaded via a `doctrine.fixture.orm` tag. This is exactly like what we did with Symfony commands.

B) There is a BC break, so, this would be 3.0. However, as long as users use the auto-registration/autoconfigure config, they won't notice the difference. Very easy update.

C) We need to Activate Travis - I've added all the config necessary

D) I removed details about the ODM support. This bundle never (or not for a long time) supported the ODM. The ODM command is inside the ODM bundle and doesn't even use this bundle (https://github.com/doctrine/DoctrineMongoDBBundle/blob/master/Command/LoadDataFixturesDoctrineODMCommand.php)

E) After merge, we can create a new recipe for 3.0 that makes the directories less deep (`src/DataFixtures/ORM` -> `src/DataFixtures`)... just because this is nicer.


.travis.yml
```diff
@@ -0,0 +1,47 @@
...
+          env: DEPENDENCIES="symfony/lts:v3"
+
+          # Latest beta release
+        - php: 7.2
+          env: DEPENDENCIES="beta"
+
+    allow_failures:
+          # Latest beta is allowed to fail.
+        - php: 7.2
+          env: DEPENDENCIES="beta"
```
alcaeus: As short as the beta phases are, I'd prefer not to have the last beta fail. Dev yes, beta should be stable IMO.

composer.json
```diff
@@ -20,7 +20,8 @@
         }
     ],
     "require": {
-        "php": ">=5.3.2",
+        "php": ">=5.5.9",
```
alcaeus: Upper bound required, please use `^5.5.9 || ^7.0`

.travis.yml
```diff
@@ -0,0 +1,47 @@
...
+    - if [ "$DEPENDENCIES" = "beta" ]; then composer config minimum-stability beta; fi;
+    - if [[ $DEPENDENCIES == *"/"* ]]; then composer require --no-update $DEPENDENCIES; fi;
+
+install:
+    # To be removed when this issue will be resolved: https://github.com/composer/composer/issues/5355
+    - if [[ "$COMPOSER_FLAGS" == *"--prefer-lowest"* ]]; then travis_retry composer update --prefer-dist --no-interaction --prefer-stable --quiet; fi
+    - travis_retry composer update ${COMPOSER_FLAGS} --prefer-dist --no-interaction
+
+script:
+    - phpunit
```
alcaeus: Does this run whatever is currently installed on travis? In that case, I'd rather see an explicit dependency on phpunit in `require-dev`.

DependencyInjection/CompilerPass/FixturesCompilerPass.php
```diff
@@ -0,0 +1,37 @@
...
+
+class FixturesCompilerPass implements CompilerPassInterface
+{
+    const FIXTURE_TAG = 'doctrine.fixture.orm';
+
+    public function process(ContainerBuilder $container)
+    {
+        $taggedServices = $container->findTaggedServiceIds(self::FIXTURE_TAG);
+
+        $references = array();
```
alcaeus: Short array syntax for new code please.

weaverryan:
Tests are passing! I've tested this on a project. This is ready to merge.

weaverryan:
Squashed!

DependencyInjection/DoctrineFixturesExtension.php
```diff
@@ -29,5 +31,8 @@ public function load(array $configs, ContainerBuilder $container)
         $loader = new XmlFileLoader($container, new FileLocator(dirname(__DIR__).'/Resources/config'));
 
         $loader->load('services.xml');
+
+        $container->registerForAutoconfiguration(ORMFixtureInterface::class)
```
alcaeus: Lovely - we shouold copy this for the MongoDBODMBundle as well.

Fixture.php
```diff
@@ -20,15 +20,6 @@
  *
  * @author Javier Eguiluz <javier.eguiluz@gmail.com>
  */
-abstract class Fixture extends AbstractFixture implements ContainerAwareInterface, DependentFixtureInterface
+abstract class Fixture extends AbstractFixture implements ORMFixtureInterface
```
alcaeus: Great change IMO. I believe dependency behavior should be added explicitly, and apart from that it also removes the `EmptyFixture` hacks 👍

Loader/SymfonyFixturesLoader.php
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * @author Ryan Weaver <ryan@knpuniversity.com>
+ */
+final class SymfonyFixturesLoader extends ContainerAwareLoader
+{
+    public function addFixture(FixtureInterface $fixture)
+    {
+        // see https://github.com/doctrine/data-fixtures/pull/274
+        // this is to give a clear error if you do not have this version
+        if (!method_exists(Loader::class, 'createFixture')) {
```
alcaeus: I'd prefer bumping the dependency version for `doctrine/data-fixtures` to `^1.3.0`, but I understand you might not want to do so due to the implied PHP 7.1 requirement for the autoloading bundle version. So, after all, I'm 👍 on this change until it makes sense to bump the PHP requirement for the bundle.

Loader/SymfonyFixturesLoader.php
```diff
@@ -0,0 +1,80 @@
...
+
+            $reflMethod = new \ReflectionMethod($dependency, '__construct');
+            foreach ($reflMethod->getParameters() as $param) {
+                if (!$param->isOptional()) {
+                    throw new \LogicException(sprintf('The getDependencies() method returned a class (%s) that has required constructor arguments. Upgrade to "doctrine/data-fixtures" version 1.3 or higher to support this.', $dependency));
+                }
+            }
+        }
+    }
+}
```
alcaeus: Missing line break at EOF.

Loader/SymfonyFixturesLoader.php
weaverryan: Exactly. It's indeed an ugly hack... and I'm sure we'll all feel comfortable removing it at *some* point in the future (hopefully not too far in the future)

DependencyInjection/DoctrineFixturesExtension.php
weaverryan: Yes! Exactly :)

weaverryan:
Nit-picks handled! Which were just to add the one missing EOL + move the beta tests to NOT allow failures (I agree on that).

Resources/doc/index.rst
```diff
@@ -212,32 +193,31 @@ Loading the Fixture Files in Order
...
-    namespace AppBundle\DataFixtures\ORM;
+    // src/DataFixtures/GroupFixtures.php
+    namespace App\DataFixtures;
     // ...
-    use AppBundle\DataFixtures\ORM\UserFixtures;
+    use App\DataFixtures\UserFixtures;
+    use Doctrine\Common\DataFixtures\DependentFixtureInterface;
 
-    class GroupFixtures extends Fixture
+    class GroupFixtures extends Fixture, implements DependentFixtureInterface;
```
alcaeus: Just noticed the comma - that probably shouldn't be there

alcaeus:
Thanks @weaverryan!

Zzortell:
:+1:

PR #210, opened 2017-11-24
hiding confusing EmptyFixture message

### LINKED ISSUE #206: [DX] Using Doctrine\Bundle\FixturesBundle\EmptyFixture is bad
Hello!

After this strange line appears in my console:
```
  > loading Doctrine\Bundle\FixturesBundle\EmptyFixture
```
I came here and read this conversation: https://github.com/doctrine/data-fixtures/pull/252

I think that if this pull-request had been accepted, it would have been great as `Doctrine\Bundle\FixturesBundle\Fixture` could implement `Doctrine\Common\DataFixtures\DependentFixtureInterface` whatever the fixture has dependencies or not.

However, the Doctrine team has ruled that no-dependency fixtures should not implement `DependentFixtureInterface`, so this bundle should not try to bypass that decision by loading an useless fixture. It can be confusing for the developer (especially if he thinks one of his dependency transformed into `EmptyFixture` ^^). It simply makes no sense to load an `EmptyFixture` if one of your fixture has no dependency. No-dependency fixtures should not implement `DependentFixtureInterface`, unless Doctrine enable `getDependencies()` returning empty array.

Moreover, the role of a Symfony bundle is to add ContainerAware ability to bridge the Doctrine data fixture library and Symfony services, not to enforce the user implementing  `DependentFixtureInterface`. It should be up to the user implementing it or not.

So I think we should remove dependency management from this bundle as well as EmptyFixture.

weaverryan:
Fixes #206 

Due to our workaround with the EmptyFixture stuff, you confusingly see an EmptyFixture class loaded. Hopefully some time we can get the PR merged into the `data-fixtures` library and remove this hack. But for now, I propose hiding the message. This fixture does *nothing*, so there is no reason for the user to know about it.

### DIFF
--- a/Command/LoadDataFixturesDoctrineCommand.php
+++ b/Command/LoadDataFixturesDoctrineCommand.php
@@ -114,6 +114,14 @@ EOT
         $purger->setPurgeMode($input->getOption('purge-with-truncate') ? ORMPurger::PURGE_MODE_TRUNCATE : ORMPurger::PURGE_MODE_DELETE);
         $executor = new ORMExecutor($em, $purger);
         $executor->setLogger(function ($message) use ($output) {
+            /*
+             * Hides the "Loading ... EmptyFixture" line, which is confusing.
+             * See the Fixture::getDependencies() method for details about this.
+             */
+            if (strpos($message, 'Doctrine\\Bundle\\FixturesBundle\\EmptyFixture') !== false) {
+                return;
+            }
+
             $output->writeln(sprintf('  <comment>></comment> <info>%s</info>', $message));
         });
         $executor->execute($fixtures, $input->getOption('append'));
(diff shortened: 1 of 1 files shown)


Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -114,6 +114,14 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $purger->setPurgeMode($input->getOption('purge-with-truncate') ? ORMPurger::PURGE_MODE_TRUNCATE : ORMPurger::PURGE_MODE_DELETE);
         $executor = new ORMExecutor($em, $purger);
         $executor->setLogger(function ($message) use ($output) {
+            /*
+             * Hides the "Loading ... EmptyFixture" line, which is confusing.
+             * See the Fixture::getDependencies() method for details about this.
+             */
+            if (strpos($message, 'Doctrine\\Bundle\\FixturesBundle\\EmptyFixture') !== false) {
```
alcaeus: Waiting for #209 would give us PHP 5.5, which would allow using the `::class` constant.

weaverryan:
Yea, let's wait for #209, then the tests will also run. I've updated to use the `::class` syntax.

weaverryan:
Closing - I've removed the EmptyFixtures in #209

PR #211, opened 2017-11-28
Data fixtures: How to use getDependencies

kamel-allab:
Source : https://github.com/doctrine/data-fixtures


alcaeus:
This is already handled in #209.

PR #212, opened 2017-11-28
Highlight code in enable bundle section

wuethrich44:
The doc looks a bit crappy without it...


alcaeus:
Looks like that slipped through the cracks. Thanks @wuethrich44!

PR #216, opened 2017-11-29
[3.0] Fix DependentFixtureInterface

### LINKED ISSUE #215: Can't load fixtures with new system
I'm tried to upgrade to version 3 of this bundle, e.g. turning my fixtures to services.
So, my starting point is some working fixtures with dependencies.
I changed all my fixtures classes to extend `Doctrine\Bundle\FixturesBundle\Fixture` instead of `Doctrine\Common\DataFixtures\AbstractFixture`.
When I try to load them, I get this error
>[LogicException]                                                                                                                                                                       
>The "App\DataFixtures\ORM\CategoryData" fixture class is trying to be loaded, but is not available. Make sure this class is defined as a service and tagged with "doctrine.fixture.orm"

I checked up with `console de:au DataFixtures` and I can see that my fixtures are listed as services. I also tried `console de:cont --show-private --tag doctrine.fixture.orm` and my fixtures are there.

The following are two example fixtures.
```php
<?php

namespace App\DataFixtures\ORM;

use Doctrine\Bundle\FixturesBundle\Fixture;
use Doctrine\Common\Persistence\ObjectManager;
use Foo\Entity\Article;

final class ArticleData extends Fixture implements DependentFixtureInterface
{
    public function load(ObjectManager $manager): void
    {
        $art = new Article('Lorem ipsum dolor', $this->getReference('category1'));
        $manager->persist($art);
        $manager->flush();
    }

    public function getDependencies(): array
    {
        return [
            CategoryData::class,
        ];
    }
}
```

```php
<?php

namespace App\DataFixtures\ORM;

use Doctrine\Bundle\FixturesBundle\Fixture;
use Doctrine\Common\Persistence\ObjectManager;
use Foo\Entity\Category;

final class CategoryData extends Fixture
{
    public function load(ObjectManager $manager): void
    {
        $cat = new Categoy('A category');
        $manager->persist($cat);
        $this->addReference('category1', $cat);
        $manager->flush();
    }
}

```

ruudk:
Fixes #215

## Problem

It seems like a racy condition with the `DependentFixtureInterface`. 

I think this is the issue:
* Every fixture is tagged with `doctrine.fixture.orm` (also the fixtures that are dependencies of others
* When the container gets created, it will do the following for the `SymfonyFixturesLoader` service:
  - Find all tagged fixture
  - For every fixture add a method call `addFixture()`
  - So it will look like this:
```php
$a = SymfonyFixturesLoader($this);
$a->addFixture(Fixture1);
$a->addFixture(Fixture2);
$a->addFixture(Fixture3);
$a->addFixture(Fixture4);
```
* When the application is booted, and you use the `SymfonyFixturesLoader` for the first time, it will execute all the `addFixture` calls
* `Doctrine\Common\DataFixtures\Loader::addFixture` will check if the fixture implements `DependentFixtureInterface`
 - When thats the case, it will call `Doctrine\Bundle\FixturesBundle\Loader\SymfonyFixturesLoader::createFixture`
 - That will then do a call to `Doctrine\Common\DataFixtures\Loader::getFixture` which then checks its internal `$this->fixtures` array to see if it exist, but it doesn't exist, because it's going to be added *later*

So I think it should be changed to do this:
* First load all fixtures into the `$this->fixtures` array
* Then foreach that array, and resolve all dependencies.

## How to test?

Add my forked repository to your `composer.json`:
```json
    "repositories": [
        {
            "type": "vcs",
            "url": "http://github.com/ruudk/DoctrineFixturesBundle"
        }
    ],
```

And then require it as follow:
```json
"require": {
    "doctrine/doctrine-fixtures-bundle": "dev-fix-dep as 3.0.0",
```

## How its solved

Instead of calling `addFixture` for every fixture, this PR will use an `addFixtures` (plural) method. It will load all the fixtures in memory, and then call `addFixture` on all of them. This will solve the dependency issues.

### DIFF
--- a/DependencyInjection/CompilerPass/FixturesCompilerPass.php
+++ b/DependencyInjection/CompilerPass/FixturesCompilerPass.php
@@ -30,8 +30,11 @@ final class FixturesCompilerPass implements CompilerPassInterface
         $definition = $container->getDefinition('doctrine.fixtures.loader');
         $taggedServices = $container->findTaggedServiceIds(self::FIXTURE_TAG);
 
+        $fixtures = [];
         foreach ($taggedServices as $serviceId => $tags) {
-            $definition->addMethodCall('addFixture', [new Reference($serviceId)]);
+            $fixtures[] = new Reference($serviceId);
         }
+
+        $definition->addMethodCall('addFixtures', [$fixtures]);
     }
 }
--- a/Loader/SymfonyFixturesLoader.php
+++ b/Loader/SymfonyFixturesLoader.php
@@ -19,8 +19,31 @@ use Symfony\Bridge\Doctrine\DataFixtures\ContainerAwareLoader;
  */
 final class SymfonyFixturesLoader extends ContainerAwareLoader
 {
+    private $loadedFixtures = [];
+
+    /**
+     * @internal
+     */
+    public function addFixtures(array $fixtures)
+    {
+        // Store all loaded fixtures so that we can resolve the dependencies correctly.
+        foreach ($fixtures as $fixture) {
+            $this->loadedFixtures[get_class($fixture)] = $fixture;
+        }
+
+        // Now load all the fixtures
+        foreach ($this->loadedFixtures as $fixture) {
+            $this->addFixture($fixture);
+        }
+    }
+
     public function addFixture(FixtureInterface $fixture)
     {
+        $class = get_class($fixture);
+        if (!isset($this->loadedFixtures[$class])) {
+            $this->loadedFixtures[$class] = $fixture;
+        }
+
         // see https://github.com/doctrine/data-fixtures/pull/274
         // this is to give a clear error if you do not have this version
         if (!method_exists(Loader::class, 'createFixture')) {
@@ -35,15 +58,18 @@ final class SymfonyFixturesLoader extends ContainerAwareLoader
      */
     protected function createFixture($class)
     {
-        try {
-            /*
-             * We don't actually need to create the fixture. We just
-             * return the one that already exists.
-             */
-            return $this->getFixture($class);
-        } catch (\InvalidArgumentException $e) {
-            throw new \LogicException(sprintf('The "%s" fixture class is trying to be loaded, but is not available. Make sure this class is defined as a service and tagged with "%s".', $class, FixturesCompilerPass::FIXTURE_TAG));
+        /*
+         * We don't actually need to create the fixture. We just
+         * return the one that already exists.
+         */
+
+        if (!isset($this->loadedFixtures[$class])) {
+            throw new \LogicException(sprintf(
+                'The "%s" fixture class is trying to be loaded, but is not available. Make sure this class is defined as a service and tagged with "%s".', $class, FixturesCompilerPass::FIXTURE_TAG
+            ));
         }
+
+        return $this->loadedFixtures[$class];
     }
 
     /**
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ install:
     # To be removed when this issue will be resolved: https://github.com/composer/composer/issues/5355
     - if [[ "$COMPOSER_FLAGS" == *"--prefer-lowest"* ]]; then travis_retry composer update --prefer-dist --no-interaction --prefer-stable --quiet; fi
     - travis_retry composer update ${COMPOSER_FLAGS} --prefer-dist --no-interaction
+    - travis_retry composer show
 
 script:
     - ./vendor/bin/simple-phpunit
--- a/Tests/IntegrationTest.php
+++ b/Tests/IntegrationTest.php
@@ -4,7 +4,6 @@ namespace Doctrine\Bundle\FixturesBundle\Tests\IntegrationTest;
 
 use Doctrine\Bundle\FixturesBundle\DependencyInjection\CompilerPass\FixturesCompilerPass;
 use Doctrine\Bundle\FixturesBundle\DoctrineFixturesBundle;
-use Doctrine\Bundle\FixturesBundle\EmptyFixture;
 use Doctrine\Bundle\FixturesBundle\Tests\Fixtures\FooBundle\DataFixtures\DependentOnRequiredConstructorArgsFixtures;
 use Doctrine\Bundle\FixturesBundle\Tests\Fixtures\FooBundle\DataFixtures\OtherFixtures;
 use Doctrine\Bundle\FixturesBundle\Tests\Fixtures\FooBundle\DataFixtures\RequiredConstructorArgsFixtures;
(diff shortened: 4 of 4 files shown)


garak:
I'm not sure that loading all fixtures in advance is an optimal solution. Can't we just load a fixture when it's needed? When calling `createFixture()` method, we could call `addFixture()` if fixture is not loaded.

ruudk:
It is working the same way as the current 3.0 version. Its just a matter of resolving the dependencies.

DependencyInjection/CompilerPass/FixturesCompilerPass.php
```diff
@@ -30,8 +30,11 @@ public function process(ContainerBuilder $container)
         $definition = $container->getDefinition('doctrine.fixtures.loader');
         $taggedServices = $container->findTaggedServiceIds(self::FIXTURE_TAG);
 
+        $fixtures = array();
```
alcaeus: Short array syntax please.

Loader/SymfonyFixturesLoader.php
```diff
@@ -19,6 +19,21 @@
  */
 final class SymfonyFixturesLoader extends ContainerAwareLoader
 {
+    private $loadedFixtures = array();
```
alcaeus: Short array syntax please.

Loader/SymfonyFixturesLoader.php
```diff
@@ -19,6 +19,21 @@
...
+
+    public function addFixtures(array $fixtures)
```
alcaeus: Technically, this would require us to release this fix as 3.1.0 due to the new public method in a class. However, since the class is final and the method is there just for internal use in a bugfix, I'd be ok releasing this as 3.0.1 if we mark the method as `@internal`, thus telling people that they shouldn't use it. Could you add that?

Loader/SymfonyFixturesLoader.php
```diff
@@ -35,15 +50,18 @@ public function addFixture(FixtureInterface $fixture)
...
-             */
-            return $this->getFixture($class);
-        } catch (\InvalidArgumentException $e) {
-            throw new \LogicException(sprintf('The "%s" fixture class is trying to be loaded, but is not available. Make sure this class is defined as a service and tagged with "%s".', $class, FixturesCompilerPass::FIXTURE_TAG));
+        /*
+         * We don't actually need to create the fixture. We just
+         * return the one that already exists.
+         */
+
+        if (!isset($this->loadedFixtures[$class])) {
```
alcaeus: This forbids using a fixture added using `addFixture` as the dependency of another fixture as it is not added to `loadedFixtures` there. `addFixture` should add the given fixture to the `loadedFixtures` array unless it's already present. In that case, there should be an exception if the fixture objects are not the same as a sanity check.

DependencyInjection/CompilerPass/FixturesCompilerPass.php
ruudk: Oops, I did this in purpose because I thought this project would require it. Will fix it :)

Loader/SymfonyFixturesLoader.php
ruudk: Fixed

DependencyInjection/CompilerPass/FixturesCompilerPass.php
ruudk: Fixed

Loader/SymfonyFixturesLoader.php
ruudk: Fixed. Added `@internal`

Loader/SymfonyFixturesLoader.php
ruudk: Fixed by adding 
```php
$class = get_class($fixture);
        if (!isset($this->loadedFixtures[$class])) {
            $this->loadedFixtures[$class] = $fixture;
        }
``` 
to `addFixture`

Would that solve it?

ruudk:
@weaverryan Added the test. Verified that it failed on master, and passes on this branch.

Loader/SymfonyFixturesLoader.php
```diff
@@ -19,8 +19,31 @@
...
+        // Now load all the fixtures
+        foreach ($this->loadedFixtures as $fixture) {
+            $this->addFixture($fixture);
+        }
+    }
+
     public function addFixture(FixtureInterface $fixture)
     {
+        $class = get_class($fixture);
+        if (!isset($this->loadedFixtures[$class])) {
```
alcaeus: It might be worth throwing an exception if `$this->loadedFixtures[$class] !== $fixture` as that indicates that something fancy is going on.

ruudk: OK I will update tomorrow and also add tests for it.

ruudk: Fixed.

ruudk: This seems to let the test fail.... will investigate :)

ruudk: @alcaeus I think the tests are failing because https://github.com/doctrine/data-fixtures/pull/274 was merged in data-fixtures 1.3.0.

How can we handle this properly? I think we should require a higher data-fixtures version?

ruudk: Since `SymfonyFixturesLoader` now depends on `SymfonyFixturesLoader::createFixture`

alcaeus: @weaverryan created a workaround to avoid bumping the dependency (which would implicitly bump the PHP requirement for the bundle as well, which they were trying to avoid). I'll take a look at it later today.

weaverryan:
@ruudk we've got test failures right now - it looks related to this bug actually. Can you take a look?

alcaeus:
I'm taking a quick look to see what's going on - I didn't expect any failures.

alcaeus:
Oh, yeah. For older versions of doctrine/fixtures the `addFixture` method actually creates a new instance, so I guess we have to remove that sanity check I suggested :(

@ruudk can you make the change? I can release this as 3.0.1 once tests are stable.

ruudk:
@alcaeus @weaverryan It's green 👊 :D

alcaeus:
Merged manually in 493961a - I excluded the changes to .travis.yml. Thanks @ruudk!

ruudk:
Ah, I left those in on purpose, because I think they are good to add. It will prevent "what the ...?" moments for others in the future... It cost me 30 min to find out that it was because the data-fixtures lib was low on olders php versions.

alcaeus:
I thought that was just put in for debugging purposes. Doesn't `composer show` pretty much repeat the output of `composer install` when no dependencies are present?

ruudk:
Ah, indeed :) Great!

ruudk:
@alcaeus It seems that you didn't merge my latest code 😱 

https://github.com/doctrine/DoctrineFixturesBundle/commit/eff2fc67133e7f9ce075bb7a4315c8f28a0bf9c7

is an older commit, i rebased it and pushed it as 
https://github.com/doctrine/DoctrineFixturesBundle/pull/216/commits/df906fd068189f99579c2bb5b4f265863095fe43

Thats the reason why master is failing!

alcaeus:
> It seems that you didn't merge my latest code

Oh wow :(

PR #222, opened 2017-12-04
Make `doctrine.fixtures.loader` public

ruudk:
I'm trying to get https://github.com/BehatExtension/DoctrineDataFixturesExtension to work with version 3.0 of this bundle.

It would help if this service was public, so that we can use this service to load all fixtures with all the dependencies.

## Related issue
https://github.com/BehatExtension/DoctrineDataFixturesExtension/issues/3

### DIFF
--- a/Resources/config/services.xml
+++ b/Resources/config/services.xml
@@ -10,7 +10,7 @@
             <tag name="console.command" command="doctrine:fixtures:load" />
         </service>
 
-        <service id="doctrine.fixtures.loader" class="Doctrine\Bundle\FixturesBundle\Loader\SymfonyFixturesLoader" public="false">
+        <service id="doctrine.fixtures.loader" class="Doctrine\Bundle\FixturesBundle\Loader\SymfonyFixturesLoader" public="true">
             <argument type="service" id="service_container" />
         </service>
     </services>


ruudk:
Tests are failing because something else. Master seems broken.

ruudk:
Master is failing because a wrong merge in https://github.com/doctrine/DoctrineFixturesBundle/pull/216#issuecomment-349077657

ruudk:
Rebased on top of https://github.com/doctrine/DoctrineFixturesBundle/pull/223 to show that tests pass :)

alcaeus:
@ruudk can you rebase on top of master please? Thanks!

ruudk:
@alcaeus Done!

alcaeus:
Thanks! 

FWIW, I can't figure out why we need to expose the loader for use in the extension you mentioned. Looking at the [service config](https://github.com/BehatExtension/DoctrineDataFixturesExtension/blob/master/src/Resources/config/services.xml#L47..L51), this might be a good opportunity to change the signature of the loader: instead of fetching services from the container, inject them properly using the service config. I don't see why we should expose services to the outside to cater to extensions that don't properly use the service container.

ruudk:
@alcaeus Hmm, interesting. I'm gonna try to see if I can achieve it in a different way.

ruudk:
My plan was to do the following: 

* When DoctrineFixturesBundle 3.0.2+ is installed, then grab all services from the Container tagged `doctrine.fixture.orm`. The issue is that the tags are only available during the building of the container. 

Since this Behat extension has no way to hook into that point (using an compiler pass), it becomes extremely hard to do this. 

Do you have another idea?

ruudk:
Or did you mean something like this: https://github.com/BehatExtension/DoctrineDataFixturesExtension/pull/4

alcaeus:
Even if you inject the loader, I wouldn't make the assumption that you want the behat extension to load the default fixtures.

ruudk:
Any recommendations? My PR is not possible because Behat (+ extensions) run in a different container than the tests.

wuethrich44:
@alcaeus I have a use case for this feature in my functional tests:

Before I execute my tests I would like to load all my fixtures in a sqlite database. I do that in the `setUp` of `KernelTestCase`:

```php
class FunctionalTestCase extends KernelTestCase
{
    private static $isDatabaseReady = false;

    protected function setUp()
    {
        if (!self::$isDatabaseReady) {
            self::bootKernel();

            $databaseInitializer = new DatabaseInitializer(self::$kernel->getContainer());
            $databaseInitializer->dropDatabase();
            $databaseInitializer->loadSchema();
            $databaseInitializer->loadFixtures();

            self::$isDatabaseReady = true;
        }
    }
...
```
Currently I have to look up all my fixtures with the `ContainerAwareLoader` and filepaths. It would be great if I could fetch the `SymfonyFixturesLoader` from the container and load my fixtures that way.

Any chance to get that merged?

alcaeus:
If you need to make a service public, it's much better to add a public alias in a special `test.xml` service file that's only loaded in your test environment. In that case, you could even make the `DatabaseInitializer` a service within `test.xml` that you can fetch from the container. That way you can also properly wire arguments instead of relying on `$this->container->get(...)`.

Given that the original use-case (using the service within the Behat extension) is no longer given (the author also suggested to change the `FixtureLoader` class to receive proper service arguments, https://github.com/BehatExtension/DoctrineDataFixturesExtension/issues/3#issuecomment-349351080) I'm closing this pull request.

wuethrich44:
@alcaeus Thanks for the hint. I could solve it with your explanation.

PR #223, opened 2017-12-04
Fix strict comparison in DependentFixtureInterface

ruudk:
See https://github.com/doctrine/DoctrineFixturesBundle/pull/216#issuecomment-349077657


alcaeus:
Thanks @ruudk!

PR #233, opened 2018-01-25
Add example for fixture autowiring

Jean85:
This is needed when upgrading a Symfony app, that will not automatically import the new default configuration for fixture classes autowiring.


alcaeus:
Thanks @Jean85!

PR #234, opened 2018-01-26
Split the framework-bundle dependency

### LINKED ISSUE #224: Avoid requiring FrameworkBundle
This bundle doesn't need symfony/framework-bundle, but it's currently requiring it.
I thinks such require should be moved to require-dev, and explicit symfony packages should be required instead:
* symfony/config
* symfony/console
* symfony/dependency-injection
* symfony/http-kernel

Taluu:
Remove the dependency on the doctrine bundle (as the only use was in the command, and none of the helpers from the `DoctrineCommand` were used), and replace the framework-bundle dependency with the config, dic and http-kernel instead.

I kept the fw-bundle dependency in dev env, because the tests need it.

Should fix #224

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,12 +25,16 @@
         "doctrine/doctrine-bundle": "^1.11|^2.0",
         "doctrine/orm": "^2.6.0",
         "symfony/doctrine-bridge": "^3.4|^4.1|^5.0",
-        "symfony/framework-bundle": "^3.4|^4.1|^5.0"
+        "symfony/config": "^3.4|^4.3|^5.0",
+        "symfony/console": "^3.4|^4.3|^5.0",
+        "symfony/dependency-injection": "^3.4|^4.3|^5.0",
+        "symfony/http-kernel": "^3.4|^4.3|^5.0"
     },
     "require-dev": {
         "doctrine/coding-standard": "^6.0",
         "phpunit/phpunit": "^7.4",
-        "symfony/phpunit-bridge": "^4.1|^5.0"
+        "symfony/phpunit-bridge": "^4.1|^5.0",
+        "symfony/framework-bundle": "^3.4|^4.0|^5.0"
     },
     "autoload": {
         "psr-4": { "Doctrine\\Bundle\\FixturesBundle\\": "" }
--- a/composer.json
+++ b/composer.json
@@ -33,8 +33,7 @@
     "require-dev": {
         "doctrine/coding-standard": "^6.0",
         "phpunit/phpunit": "^7.4",
-        "symfony/phpunit-bridge": "^4.1|^5.0",
-        "symfony/framework-bundle": "^3.4|^4.0|^5.0"
+        "symfony/phpunit-bridge": "^4.1|^5.0"
     },
     "autoload": {
         "psr-4": { "Doctrine\\Bundle\\FixturesBundle\\": "" }
--- a/Tests/IntegrationTest.php
+++ b/Tests/IntegrationTest.php
@@ -16,13 +16,10 @@ use Doctrine\Common\Persistence\ManagerRegistry;
 use PHPUnit\Framework\TestCase;
 use RuntimeException;
 use Symfony\Bridge\Doctrine\DataFixtures\ContainerAwareLoader;
-use Symfony\Bundle\FrameworkBundle\FrameworkBundle;
-use Symfony\Bundle\FrameworkBundle\Kernel\MicroKernelTrait;
 use Symfony\Component\Config\Loader\LoaderInterface;
 use Symfony\Component\DependencyInjection\Alias;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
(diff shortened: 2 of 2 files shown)


Taluu:
Hum, well it seems that the doctrine-bundle *is* needed, as it declares the `doctrine` service... So the fw-bundle can't be removed _that_ easily.

weaverryan:
@Taluu What's the status of the PR then? If we must *still* require DoctrineBundle, and DoctrineBundle requires FrameworkBundle, is there any profit in the PR?

Taluu:
IIRC, the profit would be so that this bundle does not directly require it, and wait for doctrine bundle to mark the fw-bundle as not required anymore.

Need to get into this again, for the tests.

Taluu:
Had a quick look on the bundle, as @alcaeus reminded me on this PR (:D). I guess I'll need to "break" the bc on at least the load data command, as the container, and thus the `getDoctrine` command, should be inaccessible on "modern" versions of symfony (4+)...

WDYT ?

Taluu:
Nevermind, I'll keep I guess the `DoctrineCommand` extension point.

alcaeus:
I've scheduled this for 4.0, so you're free to break BC. Just make sure to deprecate stuff first ;)

Taluu:
Nice. As not passing an argument looks like it was already deprecated, then I should be able to remove the extension on the doctrine bundle. :smile: 

But maybe in another PR, as right now the focus should be to split the fw-bundle dependency.

composer.json
```diff
@@ -22,15 +22,22 @@
...
     },
     "require-dev": {
         "doctrine/coding-standard": "^6.0",
         "phpunit/phpunit": "^7.4",
-        "symfony/phpunit-bridge": "^4.1|^5.0"
+        "symfony/phpunit-bridge": "^4.1 || ^5.0",
+        "symfony/framework-bundle": "^3.4 || ^4.0 || ^5.0"
+    },
+    "suggest": {
+        "doctrine/dbal": "To be able to use sharded connections"
```
stof: that's useless. As you depend on the ORM, which itself uses DBAL, this suggestion will never be displayed as the package will already be installed.

Thus, this suggestion is wrong. There won't be any improved feature of this bundle when installing DBAL (which is what suggestions are for). The code dealing with sharded connection in FixturesBundle is making sure that the bundle works if you use sharded connections in DoctrineBundle and DBAL. So there is nothing to suggest here.

Taluu: Yep wasn't sure if it was pertinent. @alcaeus told me that the plan was to split the bundle into the dbal bundle and the orm bundle, so maybe it will make sense to add it back as a suggestion when this happens.

I'll remove the suggestion for now.

stof:
And btw, this bundle depends on DoctrineBundle because it needs a configured ORM and DoctrineBundle is responsible for the configuration part.

Taluu:
And I think it's ready. I kept the framework-bundle as a dev dependency as its used for the tests (the integration tests, for the kernel it seems ? I'll try to see if I can rework the tests so we can do without it.

Taluu:
@stof yes that's what I saw. As I said, @alcaeus told me the plan is to split the doctrine bundle into its orm and dbal counterparts, so maybe if it's done, we'll be able to have only the orm bundle dependency.

composer.json
stof: even if we split, that won't ever be pertinent:

- splitting DoctrineBundle in 2 won't mean that the ORM will stop depending on DBAL
- the ShardedConnection logic in FixturesBundle is still not about enabling more features in FixturesBundle but about being compatible with the DBAL/DoctrineBundle features.

Taluu:
I think I succeeded in removing the (direct) dependency on the framework-bundle :tada: 

It is still used because DoctrineBundle still currently need it, but as said earlier, it should be done on the doctrine bundle (split or not).

stof:
@Taluu the ORM bundle will still depend on the DBAL bundle, as the ORM needs DBAL.
The benefit of splitting will be that the ORM bundle will be able to require the ORM (currently, DoctrineBundle has an optional dependency to ORM due to being usable in dbal-only mode) and that we will be able to have better flex recipes (as we will know whether we should configure only DBAL or ORM+DBAL thanks to bundle being separate).

composer.json
```diff
@@ -22,15 +22,18 @@
     "require": {
         "php": "^7.1",
         "doctrine/data-fixtures": "^1.3",
-        "doctrine/doctrine-bundle": "^1.11|^2.0",
+        "doctrine/doctrine-bundle": "^1.11 || ^2.0",
```
alcaeus: I'm not too keen on changing this syntax here since we use single-pipe in all other repos. Mind changing it back? I'm happy to revisit this in a separate PR.

Taluu: Even though it's the official notation (using double pipes), I changed it back.

alcaeus:
Thanks @Taluu!

PR #238, opened 2018-03-07
[DOC] Recommend constants for sharing objects between fixtures

mhujer:
I got this idea that it may be better to use constants to reference objects between fixtures. Instead of using "magic" strings (which can be renamed in one place or can contain a typo), constant creates a statically analyzable reference (so an IDE allows you to click through to the other fixture, or some other tool can verify that the reference is correct).

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -158,14 +158,16 @@ exact same object via its name:
     // ...
     class UserFixtures extends Fixture
     {
+        public const ADMIN_USER_REFERENCE = 'admin-user';
+
         public function load(ObjectManager $manager)
         {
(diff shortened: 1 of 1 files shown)


Resources/doc/index.rst
```diff
@@ -158,14 +158,16 @@ exact same object via its name:
     // ...
     class UserFixtures extends Fixture
     {
+        const ADMIN_USER_REFERENCE = 'admin-user';
```
alcaeus: `public const` please - constants have visibilities.

mhujer: fixed

alcaeus:
Thanks @mhujer!

PR #244, opened 2018-04-09
Generate URL to Doctrine

liviubalan:
Generate http://symfony.com/doc/current/doctrine.html or https://symfony.com/doc/current/doctrine.html based on current page protocol


weaverryan:
Hey! I don’t think this will work... because I think all the rst files from the Symfony core and this bundle are actually compiled independently (so these docs won’t see the /doctrine document).

However, I think we can/should just update the link to use https - no reason to link to http.

liviubalan:
Thanks for the info @weaverryan! I've changed the URL protocol to HTTPS.
Related to http/https from symfony.com website: is there a reason that there isn't a web server redirect from http to https? If you take a look on https://securityheaders.io/?q=symfony.com&followRedirects=on you will see the info "This site was served over HTTP and did not redirect to HTTPS."

PR #245, opened 2018-04-12
meta: add a tip for flex users when installing

Simperfit:
This PR adds a little tips when requiring the bundle as it has an official alias.

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -16,6 +16,11 @@ following command to download the latest stable version of this bundle:
 
     composer require --dev doctrine/doctrine-fixtures-bundle
 
+.. tip::
+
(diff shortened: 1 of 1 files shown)


Resources/doc/index.rst
```diff
@@ -16,6 +16,11 @@ following command to download the latest stable version of this bundle:
 
     composer require --dev doctrine/doctrine-fixtures-bundle
 
+.. tip::
+
+    If you are using flex you just need to do:
+    ``composer req orm-fixtures``
```
alcaeus: Should this also include `--dev` to be consistent with the `require` command above?

Simperfit: You are right. Thanks!

PR #248, opened 2018-05-11
Added service for loading data fixtures

mvmaasakkers:


Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -37,10 +36,12 @@ protected function configure()
         $this
             ->setName('doctrine:fixtures:load')
             ->setDescription('Load data fixtures to your database')
-            ->addOption('append', null, InputOption::VALUE_NONE, 'Append the data fixtures instead of deleting all data from the database first.')
+            ->addOption('append', null, InputOption::VALUE_NONE,
+                'Append the data fixtures instead of deleting all data from the database first.')
```
greg0ire: This is a violation of PSR2, either have one argument par line or put them all on a single line, same below.

Service/DoctrineFixtureService.php
```diff
@@ -0,0 +1,96 @@
...
+     * @param bool $append
+     * @return DoctrineFixtureService
+     */
+    public function setAppend($append)
+    {
+        $this->append = $append;
+
+        return $this;
+    }
+}
```
greg0ire: Please configure your IDE to always add a newline at end of file, this will make the red pictogram disappear.

Service/DoctrineFixtureService.php
```diff
@@ -0,0 +1,70 @@
...
+        $executor->setLogger($logger);
+        $executor->execute($fixtures, $this->isAppend());
+    }
+
+    public function getPurgeMode() : int
+    {
+        return $this->purgeMode;
+    }
+
+    public function setPurgeMode(int $purgeMode) : DoctrineFixtureService
```
greg0ire: I think these setters should be removed and these properties should be set through the constructor.

greg0ire:
Shouldn't there be some docs?

Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -94,25 +97,15 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-        $fixtures = $this->fixturesLoader->getFixtures($groups);
-        if (! $fixtures) {
-            $message = 'Could not find any fixture services to load';
-
-            if (! empty($groups)) {
-                $message .= sprintf(' in the groups (%s)', implode(', ', $groups));
-            }
-
-            $ui->error($message . '.');
+        $doctrineFixtureService = new DoctrineFixtureService($this->fixturesLoader, $em);
```
greg0ire: Should be injected through the constructor, this is a DIP violation.

Service/DoctrineFixtureService.php
```diff
@@ -0,0 +1,70 @@
...
+            throw new NoFixtureServicesFound('Could not find any fixture services to load.');
+        }
+        $purger = new ORMPurger($this->em);
+        $purger->setPurgeMode($this->getPurgeMode());
+        $executor = new ORMExecutor($this->em, $purger);
+        $executor->setLogger($logger);
+        $executor->execute($fixtures, $this->isAppend());
+    }
+
+    public function getPurgeMode() : int
```
greg0ire: This method is only used internally.

Service/DoctrineFixtureService.php
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public function setPurgeMode(int $purgeMode) : DoctrineFixtureService
+    {
+        $this->purgeMode = $purgeMode;
+
+        return $this;
+    }
+
+    public function isAppend() : bool
```
greg0ire: This method is only used internally

alcaeus:
I don't think this PR makes sense at this point and how it was done. The main point is to have the purger and executor injectable, not to extract the logic into a dummy service. As I've explained in https://github.com/doctrine/DoctrineFixturesBundle/issues/88#issuecomment-499841615, this is currently not possible due to how the underlying library works, but I'm hoping to fix that.

In the meantime, I'm closing this PR. Thanks for giving it a shot, @mvmaasakkers!

PR #250, opened 2018-07-03
add --fixtures PATH

rumours86:


.gitignore
```diff
@@ -1,2 +1,3 @@
 vendor/
 composer.lock
+/.idea/
```
Taluu: this is personnal ide config specific, please use your global [gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore).

Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -37,6 +38,7 @@ protected function configure()
         $this
             ->setName('doctrine:fixtures:load')
             ->setDescription('Load data fixtures to your database')
+            ->addOption('fixtures', null, InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY, 'The directory to load data fixtures from.')
```
Taluu: the value should not be optional

Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -84,9 +90,29 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+
+        $loader = new DataFixturesLoader($this->getContainer());
+        foreach ($paths as $path) {
+            if (is_dir($path)) {
+                $loader->loadFromDirectory($path);
+            } elseif (is_file($path)) {
+                $loader->loadFromFile($path);
+            }
+        }
+        $fixtures = $loader->getFixtures();
```
Taluu: with this, you can't load services fixtures, and are forcing the user to store his fixtures in `<root>/DataFixtures` or even using bundles

Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -84,9 +90,29 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         }
 
-        $fixtures = $this->fixturesLoader->getFixtures();
+        $dirOrFile = $input->getOption('fixtures');
+        if ($dirOrFile) {
+            $paths = is_array($dirOrFile) ? $dirOrFile : array($dirOrFile);
+        } else {
+            /** @var $kernel \Symfony\Component\HttpKernel\KernelInterface */
+            $kernel = $this->getApplication()->getKernel();
+            $paths = array($kernel->getRootDir().'/DataFixtures');
```
roberto910907: What do you think if we use a separate variable for storing the string?

https://github.com/doctrine/DoctrineFixturesBundle/blob/old-rewrite/Service/LoaderService.php#L39

IMHO it's better to use `sprintf` to populate the value.
https://github.com/doctrine/DoctrineFixturesBundle/blob/old-rewrite/Service/LoaderService.php#L110

Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -84,9 +90,28 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $em->getConnection()->connect($input->getOption('shard'));
         }
 
-        $fixtures = $this->fixturesLoader->getFixtures();
+        $dirOrFile = $input->getOption('fixtures');
+        if ($dirOrFile) {
+            $paths = is_array($dirOrFile) ? $dirOrFile : array($dirOrFile);
+            $loader = new DataFixturesLoader($this->getContainer());
```
Taluu: why don't you use the `$this->loader` instead ?

rumours86: because it always load all fixtures

alcaeus:
Just a quick heads up: I’m on vacation until the end of the week and will take a more in-depth look after that. After a short glance, I’m not very excited about this: the bundle uses services with a tag to find fixtures, while this proposal brings back directories. I don’t like that at all as it introduces separate systems for defining fixtures. It would be preferable to use an argument on the tag and pass that to the fixtures option instead of using directories.

.gitignore
```diff
@@ -1,2 +1,2 @@
 vendor/
-composer.lock
+composer.lock
```
alcaeus: Please avoid making changes like that - we require trailing newlines in files.

Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -84,9 +89,28 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
```
alcaeus: As previously indicated, falling back to directories is a bad idea IMO and should be changed. I've suggested using an attribute on the fixtures tag to configure the fixtures set and using that to determine fixture sets - I believe that is still the way to go. Do you want to give that a shot?

Selion05: it would be great if one fixture class could be in multiple groups. Something like `sets={"set1","set2","set3"}` or maybe passing the service ids as argument and not the path?

alcaeus:
Closing here as #260 provides a nice way of loading some fixtures instead of all of them. Thanks @rumours86 for giving this a shot!

Issue #219, opened 2017-12-01
Dynamic namespace


MassiveHiggsField:
Hi there,

i've tried to upgrade to version 3.x but i can't find a way to make it compatible with our setup. Our application has different tenants and we switch between them through an ENV variable we're setting through apache (or for cli on the shell). 
Every tenant has its own bundle with its own fixtures. Before we just do something like this:

    #~ php bin/console doctrine:fixtures:load --fixtures=src/MyApp/${tenant}Bundle/Fixtures/Entity/

Where "${tenant}" would get replaced by ant with the value of env variable TENANT, so it would be called like this:

    #~ TENANT=tenant1 ant db-reset
    #~ TENANT=tenant2 ant db-reset
    #~ TENANT=tenant3 ant db-reset

But since the "--fixtures" has been removed, i can't find a way to load different fixtures based on an environment variable.

Invis1ble:
+1, I have different fixtures for different environments, after upgrade I can't load specific fixtures.

roberto910907:
In my case I have to load fixtures per bundles and that progress inside a process is sent to a nodejs server to acomplish a real time progress in the frontend of the user. Sorry my english.

roberto910907:
I was wondering if it's possible to add the --fixtures at the command and pass it as param to [`$fixtures = $this->fixturesLoader->getFixtures()`](https://github.com/doctrine/DoctrineFixturesBundle/blob/master/Command/LoadDataFixturesDoctrineCommand.php#L93) and inside the method [`getFixtures()`](https://github.com/doctrine/data-fixtures/blob/master/lib/Doctrine/Common/DataFixtures/Loader.php#L119) filter just the fixtures in the container(in that class $this->fixtures already contains all the fixtures added in [compiler compilation](https://github.com/doctrine/DoctrineFixturesBundle/blob/master/DependencyInjection/CompilerPass/FixturesCompilerPass.php#L38) which route path coincide with the path provided in --fixture options.

What do you think?

alcaeus:
I've given this some thought - that was an oversight in the implementation that neither @weaverryan nor I thought of.

Since the service container injects all services with a specific tag into the loader, there would either have to be multiple fixture loaders (meh) or have one loader that is able to distinguish different fixture sets using a tag:

```
<!-- Will be added to the "default" set due to autoconfiguration -->
<service id="DefaultFixture" />

<!-- Will be added to the "special" set -->
<service id="...">
[...]

PR #255, opened 2018-10-09
add tag option #219 

EliasDeVos:
for issue #219 
In my Kernel.php i add the correct tag in the buildmethod since the rest is done through autoConfigure
`$container->registerForAutoconfiguration(StagingFixtures::class)
                  ->addTag('doctrine.fixture.orm', ['set' => 'staging'])
        ;`

Right know you pass --tag as a parameter to the command.
It is also not possible to change `staging` with an array because symfony gives me the following exception `A "tags" attribute must be of a scalar-type for service`

Let me know what i have to change or if things should be written different

### DIFF
--- a/Command/LoadDataFixturesDoctrineCommand.php
+++ b/Command/LoadDataFixturesDoctrineCommand.php
@@ -84,6 +84,10 @@ EOT
             $em->getConnection()->connect($input->getOption('shard'));
         }
 
+        if ($input->getOption('tag')) {
+            $this->fixturesLoader->setTag($input->getOption('tag'));
+        }
+
         $fixtures = $this->fixturesLoader->getFixtures();
         if (!$fixtures) {
             $ui->error('Could not find any fixture services to load.');
--- a/Command/LoadDataFixturesDoctrineCommand.php
+++ b/Command/LoadDataFixturesDoctrineCommand.php
@@ -38,6 +38,7 @@ class LoadDataFixturesDoctrineCommand extends DoctrineCommand
             ->setName('doctrine:fixtures:load')
             ->setDescription('Load data fixtures to your database')
             ->addOption('append', null, InputOption::VALUE_NONE, 'Append the data fixtures instead of deleting all data from the database first.')
+            ->addOption('set', null, InputOption::VALUE_OPTIONAL, 'Only fixtures with this set attribute to the doctrine.fixtures.loader tag will be loaded')
             ->addOption('em', null, InputOption::VALUE_REQUIRED, 'The entity manager to use for this command.')
             ->addOption('shard', null, InputOption::VALUE_REQUIRED, 'The shard connection to use for this command.')
             ->addOption('purge-with-truncate', null, InputOption::VALUE_NONE, 'Purge data by using a database-level TRUNCATE statement')
@@ -84,11 +85,7 @@ EOT
             $em->getConnection()->connect($input->getOption('shard'));
         }
 
-        if ($input->getOption('tag')) {
-            $this->fixturesLoader->setTag($input->getOption('tag'));
-        }
-
-        $fixtures = $this->fixturesLoader->getFixtures();
+        $fixtures = $this->fixturesLoader->getFixtures($input->getOption('set'));
         if (!$fixtures) {
             $ui->error('Could not find any fixture services to load.');
 
--- a/DependencyInjection/CompilerPass/FixturesCompilerPass.php
+++ b/DependencyInjection/CompilerPass/FixturesCompilerPass.php
@@ -21,9 +21,12 @@ final class FixturesCompilerPass implements CompilerPassInterface
 
         $fixtures = [];
         foreach ($taggedServices as $serviceId => $tags) {
-            $fixtures[] = new Reference($serviceId);
+            $fixtures[] = [
+                'fixture' => new Reference($serviceId),
+                'tags' => $tags
+            ];
         }
 
-        $definition->addMethodCall('addFixtures', [$fixtures, $taggedServices]);
+        $definition->addMethodCall('addFixtures', [$fixtures]);
     }
 }
--- a/DependencyInjection/CompilerPass/FixturesCompilerPass.php
+++ b/DependencyInjection/CompilerPass/FixturesCompilerPass.php
@@ -24,8 +24,6 @@ final class FixturesCompilerPass implements CompilerPassInterface
             $fixtures[] = new Reference($serviceId);
         }
 
-        $definition->addMethodCall('addFixtures', [$fixtures]);
-
-        $definition->addMethodCall('addFixtureTagMapping', [$taggedServices]);
+        $definition->addMethodCall('addFixtures', [$fixtures, $taggedServices]);
     }
 }
--- a/DependencyInjection/CompilerPass/FixturesCompilerPass.php
+++ b/DependencyInjection/CompilerPass/FixturesCompilerPass.php
@@ -25,5 +25,7 @@ final class FixturesCompilerPass implements CompilerPassInterface
         }
 
         $definition->addMethodCall('addFixtures', [$fixtures]);
+
+        $definition->addMethodCall('addFixtureTagMapping', [$taggedServices]);
     }
 }
--- a/Loader/SymfonyFixturesLoader.php
+++ b/Loader/SymfonyFixturesLoader.php
@@ -95,7 +95,6 @@ final class SymfonyFixturesLoader extends ContainerAwareLoader
             foreach ($fixtures as $class => $fixture) {
                 if (isset($this->setsFixtureMapping[$set][$class])) {
                     $filteredFixtures[$class] = $fixture;
-                    continue;
                 }
             }
             $fixtures = $filteredFixtures;
--- a/Loader/SymfonyFixturesLoader.php
+++ b/Loader/SymfonyFixturesLoader.php
@@ -26,19 +26,18 @@ final class SymfonyFixturesLoader extends ContainerAwareLoader
     /**
      * @internal
      */
-    public function addFixtures(array $fixtures, array $taggedServices)
+    public function addFixtures(array $fixtures)
     {
         // Store all loaded fixtures so that we can resolve the dependencies correctly.
         foreach ($fixtures as $fixture) {
             $this->loadedFixtures[get_class($fixture)] = $fixture;
+            $this->addSetsFixtureMapping(get_class($fixture['fixture']), $fixture['tags']);
         }
 
         // Now load all the fixtures
         foreach ($this->loadedFixtures as $fixture) {
             $this->addFixture($fixture);
         }
-
-        $this->addSetsFixtureMapping($taggedServices);
     }
 
     /**
@@ -82,6 +81,8 @@ final class SymfonyFixturesLoader extends ContainerAwareLoader
     /**
      * Returns the array of data fixtures to execute.
      *
+     * @param string $set
+     *
      * @return array $fixtures
      */
     public function getFixtures($set = null)
@@ -93,7 +94,7 @@ final class SymfonyFixturesLoader extends ContainerAwareLoader
             $fixtures = array_filter(
                 $fixtures,
                 function ($fixture) use ($mapping, $set) {
-                    return isset($mapping[$set]) && in_array(get_class($fixture), $mapping[$set]);
+                    return isset($mapping[$set][get_class($fixture)]) && $mapping[$set][get_class($fixture)];
                 }
             );
         }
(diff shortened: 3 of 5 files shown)


Loader/SymfonyFixturesLoader.php
```diff
@@ -72,6 +79,61 @@ protected function createFixture($class)
...
+     *
+     * @param array $services
+     *
+     * @return void
+     */
+    public function addFixtureTagMapping(array $services)
+    {
+        foreach ($services as $service => $tags) {
+            foreach ($tags as $attributes) {
+                if (key_exists('set', $attributes)) {
```
TerjeBr: Please use `array_key_exists` instead of `key_exists`.
`key_exists` is a deprecated php alias: https://bugs.php.net/bug.php?id=52806

Loader/SymfonyFixturesLoader.php
```diff
@@ -72,6 +79,61 @@ protected function createFixture($class)
...
+     * @param array $services
+     *
+     * @return void
+     */
+    public function addFixtureTagMapping(array $services)
+    {
+        foreach ($services as $service => $tags) {
+            foreach ($tags as $attributes) {
+                if (array_key_exists('set', $attributes)) {
+                    $this->fixtureTagMapping[$service] = $attributes['set'];
```
TerjeBr: Come to think of it, you will need to support multiple services for the same tag too.

So I guess you will need a nested array anyway.

EliasDeVos: yes thats what i tought, but i asked this in the description.
I can't pass an array as attribute.
Do i just pass a string and explode it ? on what character ?

TerjeBr: I was referring to `$this->fixtureTagMapping`, that can be an array of arrays.

Where do you need to have an array it the attribute? I do not follow you there.

TerjeBr: Ok, I think I get it. You cannot have multiple "set" values in one tag. You have to use multiple tags for the same service instead. I guess that is a known drawback with the tagging system in the dependency container.

TerjeBr: You do not need it. If someone wants multiple "set" tags for the same service, they can just add multiple tags for that service, with one set attibute in each tag.

Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -84,6 +84,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $em->getConnection()->connect($input->getOption('shard'));
         }
 
+        if ($input->getOption('tag')) {
+            $this->fixturesLoader->setTag($input->getOption('tag'));
```
alcaeus: This should not be stored on the loader but rather just be passed to `getFixtures` which does the filtering anyways. No need for adding the tag to the state of the object.

DependencyInjection/CompilerPass/FixturesCompilerPass.php
```diff
@@ -25,5 +25,7 @@ public function process(ContainerBuilder $container)
         }
 
         $definition->addMethodCall('addFixtures', [$fixtures]);
+
+        $definition->addMethodCall('addFixtureTagMapping', [$taggedServices]);
```
alcaeus: I don't think it's feasible to add this mapping in a second call. IMO, this should be handled in the initial call to `addFixtures` or by creating multiple calls to `addFixture` and passing a list of sets/tags the fixture belongs to as second argument.

EliasDeVos:
I changed the things you requested. i created a test but still need to test it.
I need to add it to a normal symfony project to run it right ? (new to symfony)

Loader/SymfonyFixturesLoader.php
```diff
@@ -72,6 +79,46 @@ protected function createFixture($class)
...
+     * @param array $services
+     *
+     * @return void
+     */
+    public function addSetsFixtureMapping(array $services)
+    {
+        foreach ($services as $service => $tags) {
+            foreach ($tags as $attributes) {
+                if (array_key_exists('set', $attributes)) {
+                    $this->setsFixtureMapping[$attributes['set']][] = $service;
```
TerjeBr: This can be:
```
   $this->setsFixtureMapping[$attributes['set']][$service] = true;
```

That change will make sure each service is only stored once, and later you can use `isset` instead of `in_array` (in_array is slow)

TerjeBr: As another thought, I do not think you should use the service name here.
When you look it up in the method `getFixtures` you use the class name of the service. This will normally be the same in symfony 4 projects, but it does not have to be the same. The service name could easily be something else than the class name of the object implementing the service.

To be on the safe side, I guess you need to either make sure you store the class name as a key when you store the tags, or lookup by service name instead of class name when you do the lookup.

EliasDeVos: Oke i will look into using the get_class name.
Do you happen to know where in symfony the Reference class gets replaced with its other class to use ? [here](https://github.com/doctrine/DoctrineFixturesBundle/blob/master/DependencyInjection/CompilerPass/FixturesCompilerPass.php#L24)
first its an array of References but in the addFixtures method its an array of Fixtures.
If they get created from the id in new Reference($id) will the issue you describe ever happen ?

TerjeBr: Well, during the compiler pass you have just a "Reference" to the service object. Those references are resolved when actually creating those services after all the compiler passes. So in the  service constructors and injection methods (like our "addFixtures" injection method) those references in the container being injected has been resolved to actual objects representing the services.

In symfony 4 they have the convention that the id of a service should be the fully qualified name of the class implementing the service. But it is only a convention, and before symfony 4 services usually had id's that looked something like "namespace.foo.bar.service". So the issue is very likely to happen.

TerjeBr: The problem is that the tags get passed to the method "addFixtures" with only a service id linking the tag information to the service that has been tagged. And the services themselves are passed in a different array (in the first argument `$fixtures`) with the resolved services as values. And the method "addFixtures" has no way to find a connection between a service id and those resolved service objects.

I think the solution is that the connection between service reference and tag info has to be maintained in the compiler pass before "addFixtures" is called.
May be pass as an argument just one array, where each element in the array is a 2 element array with the service and the tags info.

Loader/SymfonyFixturesLoader.php
```diff
@@ -21,14 +21,17 @@
...
+
     /**
      * @internal
      */
     public function addFixtures(array $fixtures)
     {
         // Store all loaded fixtures so that we can resolve the dependencies correctly.
         foreach ($fixtures as $fixture) {
-            $this->loadedFixtures[get_class($fixture)] = $fixture;
+            $this->loadedFixtures[get_class($fixture['fixture'])] = $fixture['fixture'];
```
TerjeBr: Instead of repeating `get_class($fixture['fixture'])]` twice, would it not be better to put it in a variable?

Loader/SymfonyFixturesLoader.php
```diff
@@ -72,6 +78,47 @@ protected function createFixture($class)
...
+    public function getFixtures($set = null)
+    {
+        $fixtures = parent::getFixtures();
+
+        if ($set) {
+            $mapping = $this->setsFixtureMapping;
+            $fixtures = array_filter(
+                $fixtures,
+                function ($fixture) use ($mapping, $set) {
+                    return isset($mapping[$set][get_class($fixture)]) && $mapping[$set][get_class($fixture)];
```
TerjeBr: You only need `isset` here. No need to check the value stored in the array, it will always be `true`.

Loader/SymfonyFixturesLoader.php
```diff
@@ -72,6 +78,47 @@ protected function createFixture($class)
...
+     */
+    public function getFixtures($set = null)
+    {
+        $fixtures = parent::getFixtures();
+
+        if ($set) {
+            $mapping = $this->setsFixtureMapping;
+            $fixtures = array_filter(
+                $fixtures,
+                function ($fixture) use ($mapping, $set) {
```
TerjeBr: Instead of passing both `$mapping` and `$set` why not just pass one variable containing `$mapping[$set]`?
That would also give you the option to return early without looping through all the fixtures if `$mapping[$set]` is empty or does not exist.

Loader/SymfonyFixturesLoader.php
```diff
@@ -72,6 +78,47 @@ protected function createFixture($class)
...
+    /**
+     * Returns the array of data fixtures to execute.
+     *
+     * @param string $set
+     *
+     * @return array $fixtures
+     */
+    public function getFixtures($set = null)
+    {
+        $fixtures = parent::getFixtures();
```
TerjeBr: Be aware that this array already has the fixture classes as keys. So you do not actually need to call `get_class` again on each fixture, because that is already stored in the array key.

I know, it is a micro optimisation, but I like code that does not do extra tasks that is not needed.

Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -38,6 +38,7 @@ protected function configure()
             ->setName('doctrine:fixtures:load')
             ->setDescription('Load data fixtures to your database')
             ->addOption('append', null, InputOption::VALUE_NONE, 'Append the data fixtures instead of deleting all data from the database first.')
+            ->addOption('set', null, InputOption::VALUE_OPTIONAL, 'Only fixtures with this set attribute to the doctrine.fixtures.loader tag will be loaded')
```
TerjeBr: Should we allow this option to be an array of sets? Or may be also allow repeated use of the set option to use more than one set?

EliasDeVos: hm maybe, i can add this if you want.
I can't really think of a use case where you would want this.
I think the main goal was to load fixtures per environment i don't know why you would wanna load multiple tags, just create a new tag then :)
for me it's not necessary.
But let me know if i need to add it

Loader/SymfonyFixturesLoader.php
```diff
@@ -72,6 +79,48 @@ protected function createFixture($class)
...
+
+        if ($set) {
+            $filteredFixtures = [];
+            foreach ($fixtures as $class => $fixture) {
+                if (isset($this->setsFixtureMapping[$set][get_class($fixture)])) {
```
TerjeBr: You should use the variable `$class` here instead of calling `get_class($fixture)`

Loader/SymfonyFixturesLoader.php
```diff
@@ -72,6 +79,48 @@ protected function createFixture($class)
...
+     *
+     * @return array $fixtures
+     */
+    public function getFixtures($set = null)
+    {
+        $fixtures = parent::getFixtures();
+
+        if ($set) {
+            $filteredFixtures = [];
+            foreach ($fixtures as $class => $fixture) {
```
TerjeBr: I see you changed from `array_map` to a `foreach` loop. You could also use `array_walk`. But -`foreach` is fine if you prefer that.

Loader/SymfonyFixturesLoader.php
EliasDeVos: my tests fail if i change this. if i test this it works. any idea why ?. i will try to fix so i can run the tests local

Loader/SymfonyFixturesLoader.php
EliasDeVos: since i am creating a new array anyway i choose foreach so i dont need to use `use`.
don't know which one is the fastest array_walk or foreach

TerjeBr: Well, you do not need to create a new array, it is ok to change the local variable `$fixtures `.

But it is up to you, this PR is fine either way.

Loader/SymfonyFixturesLoader.php
TerjeBr: I see. If one of the fixtures implements `DependentFixtureInterface`, the returned array of fixtures will not have the fixture classes as keys.

This is an unfortunate irregularity of the existing symfony code. The error is in Doctrine/Common/DataFixtures/Loader.php:308

EliasDeVos: hm i see, best to leave it like this then ?

TerjeBr: Yes, I guess so.

TerjeBr: I think you should change the loop, so that you do not assign the array key to the variable `$class`. That might confuse other developers so they make the same mistake.

If you want to keep the variable `$class`, you should assign to it on the first line of the loop:
```
    $class = get_class($fixture);
```

EliasDeVos:
@alcaeus and @weaverryan can you give an update on this pull request in relation to #256 ?
Can this one be merged ?

alcaeus:
Sorry for the delay, I haven't yet had a chance to get a good look at #256 but will check that out in the coming days and then decide how we proceed. Thank you for the patience!

EliasDeVos:
so @alcaeus don't wanna rush you, but made any decisions yet ? :)

alcaeus:
Sorry for the delay, I forgot to update the issues here. I talked to @weaverryan about this and decided that it would make sense to allow both ways - one for linking via code and one for linking via config, with the latter overriding the former. Thus, we should coordinate with Ryan and @noemiquezada to make sure we can bring these features together to use the same base (e.g. the same syntax for loading fixtures, just a different syntax for defining groups).

weaverryan:
Both approaches have been combined in #260 - and original authors' work + commits were maintained :)

PR #256, opened 2018-10-11
Enhance system to allow only executing some fixtures via an interface

noemiquezada:
This builds off of #255

It's the same idea, but you group your fixtures via an interface instead of tag. The improvement is developer experience, you stay in the Fixture class instead of needing to re-register the service in order to give it the tag.

### DIFF
--- a/Command/LoadDataFixturesDoctrineCommand.php
+++ b/Command/LoadDataFixturesDoctrineCommand.php
@@ -38,7 +38,7 @@ class LoadDataFixturesDoctrineCommand extends DoctrineCommand
             ->setName('doctrine:fixtures:load')
             ->setDescription('Load data fixtures to your database')
             ->addOption('append', null, InputOption::VALUE_NONE, 'Append the data fixtures instead of deleting all data from the database first.')
-            ->addOption('set', null, InputOption::VALUE_OPTIONAL, 'Only fixtures with this set attribute to the doctrine.fixtures.loader tag will be loaded')
+            ->addOption('groups', null, InputOption::VALUE_IS_ARRAY|InputOption::VALUE_OPTIONAL, 'Only fixtures with the following group attributes and the doctrine.fixtures.loader tag will be loaded')
             ->addOption('em', null, InputOption::VALUE_REQUIRED, 'The entity manager to use for this command.')
             ->addOption('shard', null, InputOption::VALUE_REQUIRED, 'The shard connection to use for this command.')
             ->addOption('purge-with-truncate', null, InputOption::VALUE_NONE, 'Purge data by using a database-level TRUNCATE statement')
@@ -85,7 +85,7 @@ EOT
             $em->getConnection()->connect($input->getOption('shard'));
         }
 
-        $fixtures = $this->fixturesLoader->getFixtures($input->getOption('set'));
+        $fixtures = $this->fixturesLoader->getFixtures($input->getOption('groups'));
         if (!$fixtures) {
             $ui->error('Could not find any fixture services to load.');
 
--- a/Command/LoadDataFixturesDoctrineCommand.php
+++ b/Command/LoadDataFixturesDoctrineCommand.php
@@ -38,6 +38,7 @@ class LoadDataFixturesDoctrineCommand extends DoctrineCommand
             ->setName('doctrine:fixtures:load')
             ->setDescription('Load data fixtures to your database')
             ->addOption('append', null, InputOption::VALUE_NONE, 'Append the data fixtures instead of deleting all data from the database first.')
+            ->addOption('set', null, InputOption::VALUE_OPTIONAL, 'Only fixtures with this set attribute to the doctrine.fixtures.loader tag will be loaded')
             ->addOption('em', null, InputOption::VALUE_REQUIRED, 'The entity manager to use for this command.')
             ->addOption('shard', null, InputOption::VALUE_REQUIRED, 'The shard connection to use for this command.')
             ->addOption('purge-with-truncate', null, InputOption::VALUE_NONE, 'Purge data by using a database-level TRUNCATE statement')
@@ -84,7 +85,7 @@ EOT
             $em->getConnection()->connect($input->getOption('shard'));
         }
 
-        $fixtures = $this->fixturesLoader->getFixtures();
+        $fixtures = $this->fixturesLoader->getFixtures($input->getOption('set'));
         if (!$fixtures) {
             $ui->error('Could not find any fixture services to load.');
 
--- a/DependencyInjection/CompilerPass/FixturesCompilerPass.php
+++ b/DependencyInjection/CompilerPass/FixturesCompilerPass.php
@@ -21,7 +21,10 @@ final class FixturesCompilerPass implements CompilerPassInterface
 
         $fixtures = [];
         foreach ($taggedServices as $serviceId => $tags) {
-            $fixtures[] = new Reference($serviceId);
+            $fixtures[] = [
+                'fixture' => new Reference($serviceId),
+                'tags' => $tags
+            ];
         }
 
         $definition->addMethodCall('addFixtures', [$fixtures]);
--- a/DependencyInjection/CompilerPass/FixturesCompilerPass.php
+++ b/DependencyInjection/CompilerPass/FixturesCompilerPass.php
@@ -3,6 +3,7 @@
 
 namespace Doctrine\Bundle\FixturesBundle\DependencyInjection\CompilerPass;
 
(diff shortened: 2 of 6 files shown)


weaverryan:
Hey guys!

I'm very happy to see this feature finally re-added / fixed in #255 and #256 :D. After seeing #255, I worked with @noemiquezada to change the implementation detail part. Basically, I think the having this interface is a little bit better DX than a tag. Specifically:

In #256 

* Implement the new `FixtureGroupInterface` and return the groups you want

In #255 

* Re-register the service in `services.yaml` and add the tag:

```yml
services:
    App\DataFixtures\MyFixture:
        - { name: doctrine.fixture.orm, tag: main }
```

#256 basically makes it possible to *not* need to go into YAML and re-add the service+tag. It's consistent with other places in Symfony.

However, another option is to combine #255 and #256 to allow either the tag approach OR the new interface.

Cheers!

DependencyInjection/CompilerPass/FixturesCompilerPass.php
```diff
@@ -21,9 +22,29 @@ public function process(ContainerBuilder $container)
...
+                'fixture' => new Reference($serviceId),
+                'groups' => $this->getFixtureGroups($serviceId, $container)
+            ];
+
         }
 
         $definition->addMethodCall('addFixtures', [$fixtures]);
     }
+
+    private function getFixtureGroups($service, ContainerBuilder $container) {
```
jwage: Thanks for the contribution! Can you move this trailing opening curly bracket down to the next line?

TerjeBr:
Hi.

I would very much vote for keeping a tag in the service definition as an optional way to group the fixtures. That way you can easily group a whole directory of fixtures with the same group-tag without changing each individual fixture.

Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -38,6 +38,7 @@ protected function configure()
             ->setName('doctrine:fixtures:load')
             ->setDescription('Load data fixtures to your database')
             ->addOption('append', null, InputOption::VALUE_NONE, 'Append the data fixtures instead of deleting all data from the database first.')
+            ->addOption('groups', null, InputOption::VALUE_IS_ARRAY|InputOption::VALUE_OPTIONAL, 'Only fixtures with the following group attributes and the doctrine.fixtures.loader tag will be loaded')
```
TerjeBr: InputOption::VALUE_REQUIRED would be better than InputOption::VALUE_OPTIONAL here.

It does not make sense to specify a "groups" option to the command if it has no value.

Also, would it not be better to use singular "group" instead of "groups". As I understand it, if you want more than one group you need to repeat the option anyway.

DependencyInjection/CompilerPass/FixturesCompilerPass.php
```diff
@@ -21,9 +22,30 @@ public function process(ContainerBuilder $container)
...
+                'fixture' => new Reference($serviceId),
+                'groups' => $this->getFixtureGroups($serviceId, $container)
+            ];
+
         }
 
         $definition->addMethodCall('addFixtures', [$fixtures]);
     }
+
+    private function getFixtureGroups($service, ContainerBuilder $container)
```
TerjeBr: Would it not make sense to defer this work of figuring out the groups to the SymfonyFixturesLoader?

Here you do have a lot of logic to figure out which class implements each fixture. It complicates the code, and it could also fail if a fixture service was defined by a factory method. ($def->getClass() can be different from the actual class in that case).

In the SymfonyFixturesLoader all the fixtures has been resolved to actual objects, and you could just use that object to figure out if it implemets the interface and call "getGroups" on it.

FixtureGroupInterface.php
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @author Noemi Quezada <nquezada@endertech.com>
+ */
+interface FixtureGroupInterface
+{
+    /**
+     * This method must return an array of groups
+     * on which the implementing class belongs to
+     *
+     * @return array
```
alcaeus: `@return string[]`

Loader/SymfonyFixturesLoader.php
```diff
@@ -37,6 +41,9 @@ public function addFixtures(array $fixtures)
         }
     }
 
+    /**
+     * {@inheritdoc}
```
alcaeus: This is actually useless: documentation is always inherited

Loader/SymfonyFixturesLoader.php
```diff
@@ -72,6 +79,47 @@ protected function createFixture($class)
         return $this->loadedFixtures[$class];
     }
 
+    /**
+     * Returns the array of data fixtures to execute.
+     *
+     * @param array $groups
```
alcaeus: `@param string[] $groups`

Loader/SymfonyFixturesLoader.php
```diff
@@ -72,6 +79,47 @@ protected function createFixture($class)
...
+     *
+     * @return array $fixtures
```
alcaeus: `@return FixtureInterface[]`

Loader/SymfonyFixturesLoader.php
```diff
@@ -72,6 +79,47 @@ protected function createFixture($class)
...
+        }
+
+        return $fixtures;
+    }
+
+    /**
+     * Generates an array of the groups and their fixtures
+     *
+     * @param string $className
+     * @param array $groups
```
alcaeus: `@param string[] $groups`

DependencyInjection/CompilerPass/FixturesCompilerPass.php
alcaeus: I'm conflicted. On one hand, you're right - the service instance might be a different class than the one specified in the container. On the other hand, moving it to the fixture loader would require instantiating every fixture in order to check if it's relevant, which can be avoided this way.

I'm not sure how relevant the use-case mentioned here is: could/should the Container complain if the instance a factory returns is of a different class than specified on the service?

alcaeus:
Apart from feedback above, I believe both approaches to be equally valid: as mentioned before, specifying the groups in the container via a service tag would allow assigning a group to all fixtures from a specific directory, while the solution via an interface causes less configuration overhead. However, I would like to avoid having two different ways to achieve one thing.

Having the groups in the code would be preferable, but I'd also argue that if it's not handled in the container, this should be in doctrine/migrations, not in the bundle.

DependencyInjection/CompilerPass/FixturesCompilerPass.php
TerjeBr: Instatiation of every fixture service should not be that much overhead. We are talking microseconds here in code that not very likely to be time critical any way.

I think it is much more important that the code always does the right thing.

TerjeBr:
I think it is one of Symfony's strengths that it does not try to contrain developers to "One True Way".
Having more than one way to do things I think is the right thing for symfony, when some developers would prefer one way and others would prefer another way.

I for one, would be very disappointed if I could not have a simple way to group all fixtures that where in the same directory.

Destroy666x:
I like the tag approach more, so having both would be great.

DependencyInjection/CompilerPass/FixturesCompilerPass.php
TerjeBr: Actually it is very common that a factory returns a subclass of the one declared in the container. And some subclasses could implement the FixtureGroupInterface while others do not. I think that could be a very likely scenario.

weaverryan: There's a long-standing pattern where these types of things are static - e.g. command class names, or `EventSubscriberInterface::getSusbcribedEvents()`. It makes things lazy, and instantiating objects that the user never asked to be used (e.g. fixtures in other groups) will probably lead to some WTF moments.

But, no problem! Another long-standing pattern is to add a second, non-static way of doing things: tags (this is true for both commands and event listeners). We'll do the same thing - allowing you to use that method if you need to.

weaverryan:
Both approaches have been combined in #260 - and original authors' work + commits were maintained :)

DependencyInjection/CompilerPass/FixturesCompilerPass.php
TerjeBr: Since the SymfonyFixtureLoader gets an instance of all the fixtures anyway, (in the method addFixtures), no overhead is added by deferring this work to that method. And then it can be done right.

Please see https://github.com/weaverryan/DoctrineFixturesBundle/pull/1

PR #259, opened 2018-10-27
add php 7.3 to travis test matrix

andreybolonin:


.travis.yml
```diff
@@ -17,13 +17,16 @@ matrix:
...
+        - php: 7.3
 
           # Test LTS version we support
         - php: 7.2
           env: DEPENDENCIES="symfony/lts:v3"
 
         - php: 7.2
           env: DEPENDENCIES="beta"
+        - php: 7.3
+          env: DEPENDENCIES="beta"
```
alcaeus: Instead of adding this as a separate build, we should just change the LTS and beta builds to use PHP 7.3 instead of 7.2.

PR #220, opened 2017-12-01
use SymfonyStyle for command output

OskarStark:
Refs https://github.com/doctrine/doctrine2/pull/6655

screenshot 2017-12-01 15 57 59

screenshot 2017-12-01 16 10 42

PR #260, opened 2018-11-09
Add fixture groups via a static method interface or service tags

weaverryan:
Hi guys!

This combines the nice work from #255 and #256 to add "groups" of fixtures to be loaded via a `--group=foo` flag on the command.

A fixture can be added to a group either via a `getGroups()` method on the class itself, or a `group` attribute on a service tag. This is very consistent with other patterns in Symfony - e.g. how command names can be set via a static property or as a service tag.

There is a strong IntegrationTest for this feature and I just tried it "in the wild" in a project. It works :).

Replaces #246 and #250.

Cheers!

### DIFF
--- a/Command/LoadDataFixturesDoctrineCommand.php
+++ b/Command/LoadDataFixturesDoctrineCommand.php
@@ -38,6 +38,7 @@ class LoadDataFixturesDoctrineCommand extends DoctrineCommand
             ->setName('doctrine:fixtures:load')
             ->setDescription('Load data fixtures to your database')
             ->addOption('append', null, InputOption::VALUE_NONE, 'Append the data fixtures instead of deleting all data from the database first.')
+            ->addOption('set', null, InputOption::VALUE_OPTIONAL, 'Only fixtures with this set attribute to the doctrine.fixtures.loader tag will be loaded')
             ->addOption('em', null, InputOption::VALUE_REQUIRED, 'The entity manager to use for this command.')
             ->addOption('shard', null, InputOption::VALUE_REQUIRED, 'The shard connection to use for this command.')
             ->addOption('purge-with-truncate', null, InputOption::VALUE_NONE, 'Purge data by using a database-level TRUNCATE statement')
@@ -86,7 +87,7 @@ EOT
             $em->getConnection()->connect($input->getOption('shard'));
         }
 
-        $fixtures = $this->fixturesLoader->getFixtures();
+        $fixtures = $this->fixturesLoader->getFixtures($input->getOption('set'));
         if (!$fixtures) {
             $ui->error('Could not find any fixture services to load.');
 
--- a/Command/LoadDataFixturesDoctrineCommand.php
+++ b/Command/LoadDataFixturesDoctrineCommand.php
@@ -38,7 +38,7 @@ class LoadDataFixturesDoctrineCommand extends DoctrineCommand
             ->setName('doctrine:fixtures:load')
             ->setDescription('Load data fixtures to your database')
             ->addOption('append', null, InputOption::VALUE_NONE, 'Append the data fixtures instead of deleting all data from the database first.')
-            ->addOption('set', null, InputOption::VALUE_OPTIONAL, 'Only fixtures with this set attribute to the doctrine.fixtures.loader tag will be loaded')
+            ->addOption('groups', null, InputOption::VALUE_IS_ARRAY|InputOption::VALUE_OPTIONAL, 'Only fixtures with the following group attributes and the doctrine.fixtures.loader tag will be loaded')
             ->addOption('em', null, InputOption::VALUE_REQUIRED, 'The entity manager to use for this command.')
             ->addOption('shard', null, InputOption::VALUE_REQUIRED, 'The shard connection to use for this command.')
             ->addOption('purge-with-truncate', null, InputOption::VALUE_NONE, 'Purge data by using a database-level TRUNCATE statement')
@@ -87,7 +87,7 @@ EOT
             $em->getConnection()->connect($input->getOption('shard'));
         }
 
-        $fixtures = $this->fixturesLoader->getFixtures($input->getOption('set'));
+        $fixtures = $this->fixturesLoader->getFixtures($input->getOption('groups'));
         if (!$fixtures) {
             $ui->error('Could not find any fixture services to load.');
 
--- a/Command/LoadDataFixturesDoctrineCommand.php
+++ b/Command/LoadDataFixturesDoctrineCommand.php
@@ -38,7 +38,7 @@ class LoadDataFixturesDoctrineCommand extends DoctrineCommand
             ->setName('doctrine:fixtures:load')
             ->setDescription('Load data fixtures to your database')
             ->addOption('append', null, InputOption::VALUE_NONE, 'Append the data fixtures instead of deleting all data from the database first.')
-            ->addOption('groups', null, InputOption::VALUE_IS_ARRAY|InputOption::VALUE_OPTIONAL, 'Only fixtures with the following group attributes and the doctrine.fixtures.loader tag will be loaded')
+            ->addOption('group', null, InputOption::VALUE_IS_ARRAY|InputOption::VALUE_REQUIRED, 'Only load fixtures that belong to this group')
             ->addOption('em', null, InputOption::VALUE_REQUIRED, 'The entity manager to use for this command.')
             ->addOption('shard', null, InputOption::VALUE_REQUIRED, 'The shard connection to use for this command.')
             ->addOption('purge-with-truncate', null, InputOption::VALUE_NONE, 'Purge data by using a database-level TRUNCATE statement')
@@ -57,6 +57,10 @@ By default Doctrine Data Fixtures uses DELETE statements to drop the existing ro
 If you want to use a TRUNCATE statement instead you can use the <comment>--purge-with-truncate</comment> flag:
 
   <info>php %command.full_name%</info> <comment>--purge-with-truncate</comment>
+
+To execute only fixtures that live in a certain group, use:
+
+  <info>php %command.full_name%</info> <comment>--group=group1</comment>
 
 EOT
         );
@@ -87,9 +91,16 @@ EOT
             $em->getConnection()->connect($input->getOption('shard'));
         }
 
-        $fixtures = $this->fixturesLoader->getFixtures($input->getOption('groups'));
+        $groups = $input->getOption('group');
+        $fixtures = $this->fixturesLoader->getFixtures($groups);
         if (!$fixtures) {
-            $ui->error('Could not find any fixture services to load.');
+            $message = 'Could not find any fixture services to load';
+
+            if (!empty($groups)) {
+                $message .= sprintf(' in the groups (%s)', implode(', ', $groups));
+            }
+
+            $ui->error($message . '.');
 
             return 1;
         }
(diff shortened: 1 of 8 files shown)


FixtureGroupInterface.php
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @author Noemi Quezada <nquezada@endertech.com>
+ */
+interface FixtureGroupInterface
+{
+    /**
+     * This method must return an array of groups
+     * on which the implementing class belongs to
+     *
+     * @return array
```
alcaeus: We should make this more explicit using `@return string[]`.

Loader/SymfonyFixturesLoader.php
```diff
@@ -72,6 +79,47 @@ protected function createFixture($class)
         return $this->loadedFixtures[$class];
     }
 
+    /**
+     * Returns the array of data fixtures to execute.
+     *
+     * @param array $groups
```
alcaeus: '@param string[] $groups`

Loader/SymfonyFixturesLoader.php
```diff
@@ -72,6 +79,47 @@ protected function createFixture($class)
...
+     *
+     * @return array $fixtures
```
alcaeus: `@return FixtureInterface[]`

Loader/SymfonyFixturesLoader.php
```diff
@@ -72,6 +79,47 @@ protected function createFixture($class)
...
+     */
+    public function getFixtures(array $groups = [])
+    {
+        $fixtures = parent::getFixtures();
+
+        if (!empty($groups)) {
```
alcaeus: Nitpick: use an early exit here to reduce nesting.

Loader/SymfonyFixturesLoader.php
```diff
@@ -72,6 +79,47 @@ protected function createFixture($class)
...
+        }
+
+        return $fixtures;
+    }
+
+    /**
+     * Generates an array of the groups and their fixtures
+     *
+     * @param string $className
+     * @param array $groups
```
alcaeus: `@param string[] $groups`

FixtureGroupInterface.php
```diff
@@ -0,0 +1,37 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
```
alcaeus: We no longer add license blocks to the top of files since the license in the root folder is valid without this.

TerjeBr:
@alcaeus wrote:
> @TerjeBr as for static vs. non-static, EventSubscriber interface in the event dispatcher component also has a static method with the compiler pass taking the class attribute from the service definition, so I'd say people are used to this kind of behaviour. Since I also see no point having a fixture class dynamically decide what groups to be in, I agree with handling the group interface this way.

I think there must be some misunderstanding here. I have never advocated that the method to get groups should be non-static. That was never my point. That the interface has it as a static method is totally ok with me.

What is do advocate for, is that the SymfonyFixturesLoader should call the method instead of the FixturesCompilerPass. The SymfonyFixturesLoader get passed all the fixtures any way, and there they have been resolved for you by the container to the correct classes for you. This avoids complexity in the FixturesCompilerPass, and code that might even sometimes be wrong.

See https://github.com/weaverryan/DoctrineFixturesBundle/pull/1

weaverryan:
Fixes made for the small comments!

I *did* also merge the logic from @TerjeBr. It really makes no difference where we do the logic. However, @TerjeBr did have a point about factory services, which I think is an edge case, but why not avoid it.

Should be ready to go!

TerjeBr:
This PR addesses issue https://github.com/doctrine/DoctrineFixturesBundle/issues/219

Loader/SymfonyFixturesLoader.php
```diff
@@ -72,6 +84,48 @@ protected function createFixture($class)
...
+
+        if (empty($groups)) {
+            return $fixtures;
+        }
+
+        $filteredFixtures = [];
+        foreach ($fixtures as $key => $fixture) {
+            foreach ($groups as $group) {
+                if (isset($this->groupsFixtureMapping[$group][get_class($fixture)])) {
+                    $filteredFixtures[$key] = $fixture;
```
Taluu: `yield` ?

Taluu: instead of building and returning an array, as it's gonna be foreach'd, using a generator would make more sense imo.

`yield $key => $fixture`, and replace the early return with a `yield from $fixtures;` and `return;`.

So this method would return `void`. :}

TerjeBr: Then you would also have to change the same method in the parent class `Doctrine\Common\DataFixtures\Loader`

That would be a big change across different repositories. It would also be a BC break.
But if you want to advacte for this change, I guess your best option would be to create a separate issue about it.

Taluu: Oh, didn't know it was in the parent. Carry on then. :}

alcaeus: That would be a BC break, which would have to come from doctrine/fixtures itself. Feel free to suggest this there 👍 
Also, the method wouldn't return `void` but rather `\Generator` - you can't use a `void` return value.

alcaeus:
Note: will clean up branches and release once that's done - `master` is sill `3.0.x-dev` which isn't correct.

Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -70,7 +75,9 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $em = $doctrine->getManager($input->getOption('em'));
 
         if (!$input->getOption('append')) {
-            $ui->ask('Careful, database will be purged. Do you want to continue y/N ?', false);
+            if (!$ui->confirm('Careful, database will be purged. Do you want to continue?', false)) {
```
GWillmann: Why did you change the previous options? yes/non instead of y/N?

weaverryan: It was broken before :) - if you said “n” it continued anyways. I fixed it (and also changed to the standard confirm)

sinner:
I need this feature!!! What are you waiting for approve it?

alcaeus:
> I need this feature!!! What are you waiting for approve it?

Because I’ve been sick for the past 1.5 weeks with no chance to do OSS, because I need to prepare a feature branch for 3.0 so I can merge this into master and because following discussion in 
https://github.com/doctrine/DoctrineFixturesBundle/pull/246#issuecomment-439148921 I want to actually test this to ensure it doesn’t screw something up.

Command/LoadDataFixturesDoctrineCommand.php
alcaeus: Correct, this was broken in #220 which hasn't been released yet. I've extracted the fix to #263 to have it fixed in 3.0 as well since this PR will target 3.1.

alcaeus:
@weaverryan I've extracted the confirmation and kernel fixes to the master branch, can you rebase on top of master to update the PR? I'm also dropping old PHP versions in #265, so some of the methods added in this PR can get type hints if you want to go ahead and add them 👍

TerjeBr:
Hi. Any chance of getting this in before Christmas? Or are you all busy?

alcaeus:
> Hi. Any chance of getting this in before Christmas? Or are you all busy?

Yes, and yes. I'll try to finish this up at the SymfonyCon Hackday tomorrow to get that merged. After that, I'll try to figure out a good solution for #246 and then release a new version.

FixtureGroupInterface.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Doctrine\Bundle\FixturesBundle;
+
+/**
+ * GroupsInterface needs to be implemented
```
TerjeBr: This comment is wrong now. Sould be `FixtureGroupInterface`

Loader/SymfonyFixturesLoader.php
```diff
@@ -9,10 +9,14 @@
...
 
 use Doctrine\Bundle\FixturesBundle\DependencyInjection\CompilerPass\FixturesCompilerPass;
+use Doctrine\Bundle\FixturesBundle\FixtureGroupInterface;
 use Doctrine\Common\DataFixtures\DependentFixtureInterface;
 use Doctrine\Common\DataFixtures\FixtureInterface;
-use Doctrine\Common\DataFixtures\Loader;
+use ReflectionClass;
+use RuntimeException;
 use Symfony\Bridge\Doctrine\DataFixtures\ContainerAwareLoader;
+use function array_keys;
```
TerjeBr: What is this? Why do you need use statments for these standard php functions?

Loader/SymfonyFixturesLoader.php
```diff
@@ -65,4 +83,70 @@ protected function createFixture($class)
...
+                if (isset($this->groupsFixtureMapping[$group][$fixtureClass])) {
+                    $filteredFixtures[$fixtureClass] = $fixture;
+                    continue 2;
+                }
+            }
+        }
+
+        $loadedFixtureClasses = array_keys($filteredFixtures);
+        foreach ($filteredFixtures as $fixture) {
+            $this->validateDependencies($loadedFixtureClasses, $fixture);
```
TerjeBr: You should just send the array `$filteredFixtures` as the first argument here, instead of the array keys of that array.
Then later you can use `isset` or `array_key_exists` instead of the much slower `in_array`

Loader/SymfonyFixturesLoader.php
```diff
@@ -65,4 +83,70 @@ protected function createFixture($class)
...
+    private function validateDependencies(array $loadedFixtureClasses, FixtureInterface $fixture): void
+    {
+        if (! $fixture instanceof DependentFixtureInterface) {
+            return;
+        }
+
+        $dependenciesClasses = $fixture->getDependencies();
+
+        foreach ($dependenciesClasses as $class) {
+            if (!in_array($class, $loadedFixtureClasses)) {
```
TerjeBr: Better to use `isset` or `array_key_exists`. You can do that if you pass the loaded fixture classes in the array keys instead of in the array values.

Loader/SymfonyFixturesLoader.php
alcaeus: Strictly speaking, you’d always have to call \array_keys to let PHP know you’re looking for the function from the global namespace. PHP has a fallback for it, but we decided to no longer rely on it for Doctrine. To avoid prepending every function and class from the global namespace, we import them just like we do classes.

The Doctrine Coding Standard coming up in #269 will require these for all functions used in the file.

alcaeus:
Thanks @TerjeBr, I’ll clean this up on the way back from SymfonyCon!

alcaeus:
> Hi. Any chance of getting this in before Christmas? Or are you all busy?

It took me longer than I would have liked but I just released the new version containing this feature. Merry Christmas! 🎅

borNfreee:
This is so in time!

pr

chrisciampoli:
How is this used/How do I add fixture objects to a group if I'm using a .yaml setup to define them?

alcaeus:
The supported branch configuration has not been updated (but I've pinged the people who can do that) so you had to manually switch to master to see it: https://symfony.com/doc/master/bundles/DoctrineFixturesBundle/index.html#fixture-groups-only-executing-some-fixtures

PR #262, opened 2018-11-21
Show the name of the database being purged when loading fixtures

mweimerskirch:
When working with multiple environments in Symfony and regularily reloading fixtures, it would be a useful addition to show the name of the database that's being purged before doing so. This can be a helpful last reminder to let you know if you are in the right environment and the system selected the right database.


alcaeus:
2.x is no longer maintained, please rebase this PR on top of master and update the base. Please beware that the line in question changes in #263, so you might want to wait for that to be merged before updating you're PR.

alcaeus:
Supersedes by #270.

PR #263, opened 2018-11-25
Fix confirmation prompt

alcaeus:
Extracted from #260 since the breakage occurs in 3.0 (and #260 will be 3.1).


alcaeus:
Failing tests are due to https://github.com/symfony/symfony/pull/29263 as well as deprecations in Symfony 4.2.

PR #265, opened 2018-11-25
Drop old stuff

alcaeus:
* Branch alias is now 3.1.x
* Drop PHP < 7.1
* Drop support for older versions of data-fixtures and DoctrineBundle

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "3.0.x-dev"
+            "dev-master": "3.1.x-dev"
         }
     },
     "require-dev": {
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -20,22 +20,22 @@
...
-        "doctrine/data-fixtures": "~1.0"
+        "php": "^7.1",
+        "doctrine/data-fixtures": "^1.3",
+        "doctrine/doctrine-bundle": "^1.6",
+        "symfony/doctrine-bridge": "~3.4|^4.0",
+        "symfony/framework-bundle": "^3.4|^4.0"
+    },
+    "require-dev": {
+        "phpunit/phpunit": "^7.4",
+        "symfony/phpunit-bridge": "^3.4|^4.0"
```
stof: For the bridge, you can even require `^4.1` all the time

PR #269, opened 2018-12-03
[WIP] Apply Doctrine Coding Standard

carusogabriel:
Missing:
- [ ] Add CI
- [ ] Add `composer.lock`
- [ ] Fix BC rules


Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -32,7 +33,7 @@ public function __construct(SymfonyFixturesLoader $fixturesLoader)
         $this->fixturesLoader = $fixturesLoader;
     }
 
-    protected function configure()
+    protected function configure() : void
```
alcaeus: This is a BC break: the class is not final and an extending class may override this method.

Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -66,23 +67,23 @@ protected function configure()
         );
     }
 
-    protected function execute(InputInterface $input, OutputInterface $output)
+    protected function execute(InputInterface $input, OutputInterface $output) : void
```
alcaeus: BC break here as well.

Loader/SymfonyFixturesLoader.php
```diff
@@ -1,41 +1,39 @@
...
 use function array_values;
+use function get_class;
+use function sprintf;
 
-/**
- * @author Ryan Weaver <ryan@knpuniversity.com>
- */
 final class SymfonyFixturesLoader extends ContainerAwareLoader
 {
+    /** @var array */
```
alcaeus: Wen can use `FixtureInterface[]` here.

Loader/SymfonyFixturesLoader.php
```diff
@@ -1,41 +1,39 @@
...
     private $loadedFixtures = [];
 
+    /** @var array<string, array<string, bool>> */
```
alcaeus: Not sure this is correct syntax, but it's the correct data structure.

phpcs.xml.dist
```diff
@@ -0,0 +1,34 @@
...
+    <arg name="basepath" value="."/>
+    <arg name="extensions" value="php"/>
+    <arg name="parallel" value="45"/>
+    <arg name="cache" value=".phpcs-cache"/>
+    <arg name="colors"/>
+
+     <!-- Ignore warnings and show progress of the run -->
+    <arg value="np"/>
+
+    <file>Command</file>
```
alcaeus: Instead of explicitly listing files, we should include `.` and exclude the `vendor` folder. Otherwise we won't check the files that are in the root folder of the bundle.

carusogabriel:
@alcaeus Thanks for the review. I'll only be able to work on this patch after December 17th. If, for some reason, this patch is blocking others, feel free to finish it :)

alcaeus:
@carusogabriel thanks for the info, I'll probably finish this before then. Thanks for getting us started here 👍

PR #270, opened 2018-12-11
Show the name of the database being purged when loading fixtures

alcaeus:
Supersedes #262. Thanks @mweimerskirch!

### DIFF
--- a/Command/LoadDataFixturesDoctrineCommand.php
+++ b/Command/LoadDataFixturesDoctrineCommand.php
@@ -78,7 +78,7 @@ EOT
         $em       = $doctrine->getManager($input->getOption('em'));
 
         if (! $input->getOption('append')) {
-            if (! $ui->confirm('Careful, database will be purged. Do you want to continue?', ! $input->isInteractive())) {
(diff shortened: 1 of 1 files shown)


Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -78,7 +78,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $em       = $doctrine->getManager($input->getOption('em'));
 
         if (! $input->getOption('append')) {
-            if (! $ui->confirm('Careful, database will be purged. Do you want to continue?', ! $input->isInteractive())) {
+            if (! $ui->confirm(sprintf('Careful, database %s will be purged. Do you want to continue?', $em->getConnection()->getDatabase()), ! $input->isInteractive())) {
```
javiereguiluz: Not sure what's Doctrine's policy, but in Symfony exception messages we wrap user-generated info with quotes to avoid misunderstandings: `database %s will be purged` -> `database "%s" will be purged`.

alcaeus: We don’t have a policy, but it makes sense to do so. Will update it 👍

PR #275, opened 2019-03-26
Updated docs to match the style of the other Symfony docs

javiereguiluz:
Most changes are cosmetic ... but I also promoted Symfony4/Flex more in the docs.

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -9,22 +9,24 @@ supports the `ORM`_ (MySQL, PostgreSQL, SQLite, etc.).
 Installation
 ------------
 
-Open a command console, enter your project directory and run the
-following command to download the latest stable version of this bundle:
+In Symfony 4 or higher applications that use `Symfony Flex`_, open a command
+console, enter your project directory and run the following command:
 
-.. code-block:: bash
+.. code-block:: terminal
 
-    composer require --dev doctrine/doctrine-fixtures-bundle
+    $ composer require --dev orm-fixtures
 
-.. tip::
+That's all! You can skip to the next section and start writing fixtures.
 
-    If you are using flex you just need to do:
-    ``composer req --dev orm-fixtures``
+In Symfony 3 applications (or when not using Symfony Flex), run this other
+command instead:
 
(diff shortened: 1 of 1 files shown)


Resources/doc/index.rst
```diff
@@ -9,22 +9,24 @@ supports the `ORM`_ (MySQL, PostgreSQL, SQLite, etc.).
...
+
+.. code-block:: terminal
 
-If you're *not* using Symfony Flex (i.e. Symfony 3 and lower), you will
-also need to enable the bundle in your ``AppKernel`` class:
+    $ composer require --dev doctrine/doctrine-fixtures-bundle
 
-.. code-block:: php
+You will also need to enable the bundle. In Symfony 3 and earlier applications,
+update the ``AppKernel`` class::
```
alcaeus: Is the duplicated colon a mistake or does it serve a purpose?

javiereguiluz: The double colon means: highlight the following code block using the default language (which is `php` in Symfony Docs). We do this to avoid adding `.. code-block:: php` continuously.

When parsing it, the two colons are transformed into a single colon followed by the highlighted code.

javiereguiluz:
I rebased to 3.1. Thanks.

Resources/doc/index.rst
```diff
@@ -9,22 +9,24 @@ supports the `ORM`_ (MySQL, PostgreSQL, SQLite, etc.).
 Installation
 ------------
 
-Open a command console, enter your project directory and run the
-following command to download the latest stable version of this bundle:
+In Symfony 4 or higher applications that use `Symfony Flex`_, open a command
+console, enter your project directory and run the following command:
```
greg0ire: This sentence sounds a bit weird to my ears… how about this instead :

Applications that use Symfony 4 or higher along with Symfony Flex`_, open a terminal…

Resources/doc/index.rst
alcaeus: Thanks for explaining 👍

Resources/doc/index.rst
javiereguiluz: I've made some minor rewords. Thanks.

PR #280, opened 2019-05-27
Fix deprecation on load fixtures command

yceruto:
```
  1x: The "Doctrine\Bundle\FixturesBundle\Command\LoadDataFixturesDoctrineCommand" constructor expects a "Doctrine\Common\Persistence\ManagerRegistry" instance as first argument, not passing it will throw a \TypeError in DoctrineBundle 2.0.
    1x in Application::run from Symfony\Bundle\FrameworkBundle\Console
  1x: The "Doctrine\Bundle\DoctrineBundle\Command\DoctrineCommand::getContainer()" method is deprecated and will be removed in DoctrineBundle 2.0.
    1x in Application::run from Symfony\Bundle\FrameworkBundle\Console
```

### DIFF
--- a/Command/LoadDataFixturesDoctrineCommand.php
+++ b/Command/LoadDataFixturesDoctrineCommand.php
@@ -15,8 +15,10 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
+use const E_USER_DEPRECATED;
 use function implode;
 use function sprintf;
+use function trigger_error;
 
 /**
  * Load data fixtures from bundles.
@@ -26,9 +28,17 @@ class LoadDataFixturesDoctrineCommand extends DoctrineCommand
     /** @var SymfonyFixturesLoader */
     private $fixturesLoader;
 
-    public function __construct(SymfonyFixturesLoader $fixturesLoader)
+    public function __construct(SymfonyFixturesLoader $fixturesLoader, ?ManagerRegistry $doctrine = null)
     {
-        parent::__construct();
+        if ($doctrine === null) {
+            @trigger_error(sprintf(
+                'The "%s" constructor expects a "%s" instance as second argument, not passing it will throw a \TypeError in DoctrineFixturesBundle 4.0.',
+                static::class,
+                ManagerRegistry::class
+            ), E_USER_DEPRECATED);
+        }
+
+        parent::__construct($doctrine);
 
         $this->fixturesLoader = $fixturesLoader;
     }
--- a/Resources/config/services.xml
+++ b/Resources/config/services.xml
@@ -7,6 +7,7 @@
         <!-- commands -->
         <service id="doctrine.fixtures_load_command" class="Doctrine\Bundle\FixturesBundle\Command\LoadDataFixturesDoctrineCommand">
             <argument type="service" id="doctrine.fixtures.loader" />
+            <argument type="service" id="doctrine" />
             <tag name="console.command" command="doctrine:fixtures:load" />
         </service>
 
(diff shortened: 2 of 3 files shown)


Resources/config/services.xml
```diff
@@ -7,6 +7,7 @@
         <!-- commands -->
         <service id="doctrine.fixtures_load_command" class="Doctrine\Bundle\FixturesBundle\Command\LoadDataFixturesDoctrineCommand">
             <argument type="service" id="doctrine.fixtures.loader" />
+            <argument type="service" id="doctrine" on-invalid="null" />
```
alcaeus:
```suggestion
            <argument type="service" id="doctrine" />
```

Since this bundle has a hard dependency on doctrine-bundle, not having a `doctrine` service should not silently fail but rather fail hard. Was there a particular reason why you added it?

yceruto: Yes, because integration tests don't include it, but I think you are right and tests should be fixed.

yceruto: done.

Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -26,9 +26,9 @@ class LoadDataFixturesDoctrineCommand extends DoctrineCommand
     /** @var SymfonyFixturesLoader */
     private $fixturesLoader;
 
-    public function __construct(SymfonyFixturesLoader $fixturesLoader)
+    public function __construct(SymfonyFixturesLoader $fixturesLoader, ?ManagerRegistry $doctrine = null)
     {
-        parent::__construct();
+        parent::__construct($doctrine);
```
alcaeus: Coming to think of it, we should throw a deprecation notice similar to the one in DoctrineBundle (see https://github.com/doctrine/DoctrineBundle/blob/eb068c210a8c735c087ac93f9754b2798aec00a8/Command/DoctrineCommand.php#L31..L37).

yceruto: Yeah, makes sense, I'll do it ASAP, thanks!

yceruto: Done.

alcaeus:
Thanks @yceruto!

yceruto:
@alcaeus Is there a plan to release a new version to get rid of this warning everywhere? Thank you.

alcaeus:
I’ll take time to do some housekeeping in the bundle next week 👍

alcaeus:
3.2.0 is released. Thanks @yceruto!

Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -26,9 +28,17 @@ class LoadDataFixturesDoctrineCommand extends DoctrineCommand
...
-        parent::__construct();
+        if ($doctrine === null) {
+            @trigger_error(sprintf(
+                'The "%s" constructor expects a "%s" instance as second argument, not passing it will throw a \TypeError in DoctrineFixturesBundle 4.0.',
+                static::class,
+                ManagerRegistry::class
+            ), E_USER_DEPRECATED);
+        }
+
+        parent::__construct($doctrine);
```
fsevestre: Hello,

The Symfony `Command` constructor argument is waiting for a string and not an object. `https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Console/Command/Command.php#L69`

On previous SF versions (without PHP 7 type), it breaks here (https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/Console/Command/Command.php#L660), which is calls by the `setName`.

Running `bin/console --env=test` for example returns
```
In Command.php line 659:
                                                                        
  Warning: preg_match() expects parameter 2 to be string, object given
```

alcaeus: Please use 3.2.2 which fixes this. Thanks!

PR #283, opened 2019-06-07
avoid deprecation notice from `LoadDataFixturesDoctrineCommand`

craue:
To get rid of the deprecation notice regarding `Doctrine\Bundle\DoctrineBundle\Command\DoctrineCommand::getContainer()`. This was missing in #280.

### DIFF
--- a/Command/LoadDataFixturesDoctrineCommand.php
+++ b/Command/LoadDataFixturesDoctrineCommand.php
@@ -83,9 +83,7 @@ EOT
     {
         $ui = new SymfonyStyle($input, $output);
 
-        /** @var ManagerRegistry $doctrine */
-        $doctrine = $this->getContainer()->get('doctrine');
-        $em       = $doctrine->getManager($input->getOption('em'));
+        $em = $this->getDoctrine()->getManager($input->getOption('em'));
 
         if (! $input->getOption('append')) {
             if (! $ui->confirm(sprintf('Careful, database "%s" will be purged. Do you want to continue?', $em->getConnection()->getDatabase()), ! $input->isInteractive())) {


Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -84,7 +84,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $ui = new SymfonyStyle($input, $output);
 
         /** @var ManagerRegistry $doctrine */
-        $doctrine = $this->getContainer()->get('doctrine');
+        $doctrine = $this->getDoctrine();
```
greg0ire: I think your change makes line 86 useless, as well as this one time variable. This should probably be changed to `$em = $this->getDoctrine()->getManager($input->getOption('em'))`

yceruto:
Thanks @craue !

All these changes (included https://github.com/doctrine/DoctrineFixturesBundle/pull/280) to get rid of deprecations make this bundle now incompatible with DoctrineBundle 1.9 and prior versions. For example with DoctrineBundle 1.9 you'll get an error as doctrine instance is not the expected command name and `getDoctrine()` method does not exist.

Command/LoadDataFixturesDoctrineCommand.php
craue: True. Done.

craue:
@yceruto, I wonder why this one wasn't triggered while running tests on Travis for #280.

alcaeus:
> @yceruto, I wonder why this one wasn't triggered while running tests on Travis for #280.

The tests are quite rudimentary, so there's a good chance the code is not being hit.

> All these changes (included #280) to get rid of deprecations make this bundle now incompatible with DoctrineBundle 1.9 and prior versions. For example with DoctrineBundle 1.9 you'll get an error as doctrine instance is not the expected command name and `getDoctrine()` method does not exist.

#280 should be fine, as passing an extra argument to the constructor isn't a problem. However, this will make it incompatible with DoctrineBundle 1.9. I see two options here:
1. make the call to `getDoctrine` conditional: if the method exists, use the new logic, otherwise use the old.
2. Bump the minimum version of doctrine/doctrine-bundle to 1.10.x where the `getDoctrine` exists.

The first option allows us to fix this in a patch release, while the second option definitely requires a new minor release. I'd suggest using option  and cleaning up the dependency in the next version.

yceruto:
> #280 should be fine, as passing an extra argument to the constructor isn't a problem.

In #280 the `$doctrine` instance is passed to the `parent::__construct()` method:
https://github.com/doctrine/DoctrineFixturesBundle/blob/f5a3f1193c1b00173d3d3c16fd9eaeb26aa36edb/Command/LoadDataFixturesDoctrineCommand.php#L41

thus, it [will be validated](https://github.com/symfony/symfony/blob/9d7e9fcac9b647e6b6df25bc35020c5b5734273c/src/Symfony/Component/Console/Command/Command.php#L415) as [command name](https://github.com/symfony/symfony/blob/9d7e9fcac9b647e6b6df25bc35020c5b5734273c/src/Symfony/Component/Console/Command/Command.php#L74), which will fail on [`preg_match()`](https://github.com/symfony/symfony/blob/9d7e9fcac9b647e6b6df25bc35020c5b5734273c/src/Symfony/Component/Console/Command/Command.php#L650).

alcaeus:
Oh, yeah. I completely missed that. I’ll fix it later today in a patch release.

alcaeus:
Superseded by #285. Thanks for bringing this up, @craue!

PR #285, opened 2019-06-12
Fix deprecations in fixture command

alcaeus:
Supersedes #283.

This PR:
* fixes an invalid call to `DoctrineCommand::__construct` on lower DoctrineBundle versions
* removes a call to the deprecated `getContainer()` method if the `getDoctrine()` method is present.

Since this is a bugfix release for 3.2, the dependency to DoctrineBundle can't really be bumped, as it will simply make this bugfix ineligible during version resolution for users of older DoctrineBundle versions. Thus, we work with `method_exists` to ensure the calls are handled correctly, removing this workaround in the next minor release where we'll properly bump the DoctrineBundle dependency.

### DIFF
--- a/Command/LoadDataFixturesDoctrineCommand.php
+++ b/Command/LoadDataFixturesDoctrineCommand.php
@@ -17,6 +17,7 @@ use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use const E_USER_DEPRECATED;
 use function implode;
+use function method_exists;
 use function sprintf;
 use function trigger_error;
 
@@ -38,7 +39,8 @@ class LoadDataFixturesDoctrineCommand extends DoctrineCommand
             ), E_USER_DEPRECATED);
         }
 
-        parent::__construct($doctrine);
+        // @todo The method_exists call can be removed once the DoctrineBundle dependency has been bumped to at least 1.10
+        parent::__construct(method_exists($this, 'getDoctrine') ? $doctrine : null);
 
         $this->fixturesLoader = $fixturesLoader;
     }
--- a/Command/LoadDataFixturesDoctrineCommand.php
+++ b/Command/LoadDataFixturesDoctrineCommand.php
@@ -85,9 +85,15 @@ EOT
     {
         $ui = new SymfonyStyle($input, $output);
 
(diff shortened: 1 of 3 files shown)


Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -38,7 +39,8 @@ public function __construct(SymfonyFixturesLoader $fixturesLoader, ?ManagerRegis
             ), E_USER_DEPRECATED);
         }
 
-        parent::__construct($doctrine);
+        // @todo The method_exists call can be removed once the DoctrineBundle dependency has been bumped to at least 1.10
```
Ocramius: Can't you bump the dependency and release a new minor instead?

alcaeus: That would still leave a broken version for people on older versions of DoctrineBundle, which I'd like to avoid.

Ocramius: Fair enough, as long as this diff does get changed when merging up 👍

Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -83,9 +85,15 @@ protected function execute(InputInterface $input, OutputInterface $output)
     {
         $ui = new SymfonyStyle($input, $output);
 
-        /** @var ManagerRegistry $doctrine */
-        $doctrine = $this->getContainer()->get('doctrine');
-        $em       = $doctrine->getManager($input->getOption('em'));
+        // @todo The method_exists call can be removed once the DoctrineBundle dependency has been bumped to at least 1.10
+        if (method_exists($this, 'getDotrine')) {
```
craue: `getDotrine`? 😏

alcaeus: Oh come on...

Ocramius: Ouch... also didn't see this :S

alcaeus: Fixed 🙄

alcaeus: @Ocramius I'm still working on tests that manage to get to that point in the functionality. Will most likely end with completely changing the command in 4.0...

PR #292, opened 2019-08-23
WIP: Add "Default" group to fixtures that have no group

### LINKED ISSUE #291: Allow to load fixtures that have no group
Hi,

When starting to use the groups, I needed to load all fixtures BUT not the one I've put in a group (that are specific fixtures that are used for some tests). 

For instance  when using the serialization groups in JMS Serializer, all objects that have no groups are put in a group `Default`. 
See : https://jmsyst.com/libs/serializer/master/cookbook/exclusion_strategies#overriding-groups-of-deeper-branches-of-the-graph

It would work like this:
* `doctrine:fixtures:load --group=Default`: load all fixtures that have no group
* `doctrine:fixtures:load --group=Default --group=User`: load all fixtures that have no group and the User group
* `doctrine:fixtures:load --group=User`: load only the User group

WDYT ?

magnetik:
Should fix #291

### DIFF
--- a/Loader/SymfonyFixturesLoader.php
+++ b/Loader/SymfonyFixturesLoader.php
@@ -55,6 +55,8 @@ final class SymfonyFixturesLoader extends ContainerAwareLoader
 
         if ($fixture instanceof FixtureGroupInterface) {
             $this->addGroupsFixtureMapping($class, $fixture::getGroups());
+        } else {
+            $this->addGroupsFixtureMapping($class, ['Default']);
         }
 
         parent::addFixture($fixture);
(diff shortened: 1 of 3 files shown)


SenseException (changes requested):
Thank you for your contribution. Before we can start a review, there is at least one missing test for your feature you mentioned in #291.

You currently only test `doctrine:fixtures:load --group=Default`, but not the other use case where you use `Default` a given fixture group. Please add a test for this case too.

This addition will also need new entries in the documentation.

PReimers:
@magnetik any news regarding the missing test here?

derrabus:
After 5 years, I don't believe we'll get an answer here anymore. Let's close.

PR #295, opened 2019-09-26
Update index.rst

abdounikarim:
In Symfony4, Flex add automatically the config in config/bundles.php

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -16,7 +16,9 @@ console, enter your project directory and run the following command:
 
     $ composer require --dev orm-fixtures
 
-That's all! You can skip to the next section and start writing fixtures.
+Starting from Symfony 4.0, Flex should be used by default and register the
+bundle for you, and in that case you can skip to the next section and start
+writing fixtures.
 
 In Symfony 3 applications (or when not using Symfony Flex), run this other
 command instead:
@@ -36,14 +38,6 @@ update the ``AppKernel`` class::
         // ...
         $bundles[] = new Doctrine\Bundle\FixturesBundle\DoctrineFixturesBundle();
     }
-
-In Symfony 4 applications, update the ``config/bundles.php`` file::
-
-    // config/bundles.php
-    return [
-        // ...
-        Doctrine\Bundle\FixturesBundle\DoctrineFixturesBundle::class => ['dev' => true, 'test' => true],
-    ];
 
 Writing Fixtures
 ----------------


greg0ire:
Not everyone uses flex though: https://twitter.com/Ocramius/status/1175051246601940992

Maybe keep this block but add a comment about how it should not be need on an application that has flex?

nicolas-grekas:
`config/bundles.php` is typically from Flex, ppl that opt-out from it while keeping the directory layout know what they're doing.

greg0ire:
Ok but it still reads a bit weird to me: "You will also need to enable the bundle. In Symfony 3 and earlier applications, update the ``AppKernel`` class:[…]". The reader might wonder why there is nothing about Symfony4. So I would drop the block, but add a line that would say "In Symfony 4, Flex should do it for you."

abdounikarim:
@greg0ire 
Is not this message clear enough ?
**That's all! You can skip to the next section and start writing fixtures.**

Or do you want to have this ?
**That's all! You can skip to the next section and start writing fixtures. You have nothing else to do, Symfony Flex do the work for you**

greg0ire:
I'd use `In Symfony 4 applications, Flex should register the bundle for you, you can skip to the next section and start writing fixtures.`

SenseException:
Flex can also be used in Symfony 3.4, so I wouldn't connect that information to a Symfony version.

https://symfony.com/doc/3.4/setup/flex.html

> Starting from Symfony 4.0, Flex should be used by default, but it is still optional.

greg0ire:
How about this:

> Starting from Symfony 4.0, Flex should be used by default and register the bundle for you, and in that case you can skip to the next section and start writing fixtures

SenseException:
That sounds good.

@abdounikarim: Can you please use @greg0ire suggestion in your PR?

abdounikarim:
@SenseException actually, no 😛Because, like you said, Flex is also in Symfony 3.4. 
Is this correct for you : 
Starting from Symfony 3.4, Flex should be used by default and register the bundle for you, and in that case you can skip to the next section and start writing fixtures.

abdounikarim:
@SenseException if Symfony 3.3 is no longer supported, could we also remove informations about Symfony 3 ? And just leave infos about Symfony 3.4 and 4 ?

SenseException:
Sounds like a plan.

@greg0ire What do you think about that?

greg0ire:
I don't know… how do "infos about Symfony 3" differ in any ways from "infos about Symfony 3.4"? If Flex is not the default in Symfony 3.4, then they should be the same to me.

Resources/doc/index.rst
```diff
@@ -16,7 +16,7 @@ console, enter your project directory and run the following command:
 
     $ composer require --dev orm-fixtures
 
-That's all! You can skip to the next section and start writing fixtures.
+Starting from Symfony 4.0, Flex should be used by default and register the bundle for you, and in that case you can skip to the next section and start writing fixtures
```
greg0ire: Please comply with the sf docs standards:

> [Each line should break approximately after the first word that crosses the 72nd character (so most lines end up being 72-78 characters);](http://symfony.com/doc/current/contributing/documentation/standards.html)

greg0ire:
Please kindly squash your commits together. If you don't, we'll try to remember to do it for you but it's best if you save us this trouble.

How to do that?

1. `git rebase -i origin/master`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.
2. A window will show up with many lines, replace `pick` with `fixup` on every line but the first one
3. Close your editor, git should do its magic, and you should end up with one commit
4. Use `git push --force` to overwrite what you already push. Don't forget the `--force` option otherwise git will try to merge both things together.

greg0ire:
Got an additional approval from Nicolas Grekas at the Hacktoberfest, so I merged.

Issue #301, opened 2019-12-26
group delivered as string but not array


Codealist:
`
[2019-12-26 12:12:03] console.ERROR: Error thrown while running command "doctrine:fixtures:load --em=platform --append --group=platform". Message: "Argument 1 passed to Doctrine\Bundle\FixturesBundle\Loader\SymfonyFixturesLoader::getFixtures() must be of the type array, string given, called in /opt/project/backend/vendor/doctrine/doctrine-fixtures-bundle/Command/LoadDataFixturesDoctrineCommand.php on line 116" {"exception":"[object] (TypeError(code: 0): Argument 1 passed to Doctrine\\Bundle\\FixturesBundle\\Loader\\SymfonyFixturesLoader::getFixtures() must be of the type array, string given, called in /opt/project/backend/vendor/doctrine/doctrine-fixtures-bundle/Command/LoadDataFixturesDoctrineCommand.php on line 116 at /opt/project/backend/vendor/doctrine/doctrine-fixtures-bundle/Loader/SymfonyFixturesLoader.php:93)","command":"doctrine:fixtures:load --em=platform --append --group=platform","message":"Argument 1 passed to Doctrine\\Bundle\\FixturesBundle\\Loader\\SymfonyFixturesLoader::getFixtures() must be of the type array, string given, called in /opt/project/backend/vendor/doctrine/doctrine-fixtures-bundle/Command/LoadDataFixturesDoctrineCommand.php on line 116"} []
`

The main inconsistency here is actually in lines:
```
$groups   = $input->getOption('group');
$fixtures = $this->fixturesLoader->getFixtures($groups);
```
`$input->getOption` returns mixed
However `::getFixtures` method allows to pass groups only as an array
```
public function getFixtures(array $groups = []) : array
```
It is a quite simple fix, i will prepare a pull request for that in a minute

PR #302, opened 2019-12-26
Safe array cast to be sure that we are consistent in interfaces …

idoraquel:
…on a way from InputInterface::getOption() to ::getFixtures()

Fixes https://github.com/doctrine/DoctrineFixturesBundle/issues/301

### DIFF
--- a/Command/LoadDataFixturesDoctrineCommand.php
+++ b/Command/LoadDataFixturesDoctrineCommand.php
@@ -105,7 +105,7 @@ EOT
             $em->getConnection()->connect($input->getOption('shard'));
         }
 
-        $groups   = $input->getOption('group');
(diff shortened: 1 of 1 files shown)


greg0ire:
This was introduced in https://github.com/doctrine/DoctrineFixturesBundle/pull/260 . It looks funny to me that we have to do this… isn't there a way to consistently have an array when calling `getOption`?

greg0ire:
Also, looks like the build is broken on at least master.

alcaeus:
@Codealist why is this necessary? From looking at the code, `getOption` should always return an array, even if only one option was given. Can you provide a test case that fails with the current code but passes with yours? Thanks!

PR #303, opened 2019-12-30
Fix build

greg0ire:


greg0ire:
:thinking: the 7.4 job fails, maybe it doesn't pull the same deps as the other jobs?

greg0ire:
I now believe this has to do with the data-fixtures lib

greg0ire:
I made a PR to address this: https://github.com/doctrine/data-fixtures/pull/333

greg0ire:
Ok so now it's slightly better… but the version of the lib I'm requiring requires php 7.2… should I drop 7.1?

Issue #304, opened 2020-03-10
Composer deprecation notices


adrienfr:
Hi,

Since composer 1.10 release ([here](https://github.com/composer/composer/releases/tag/1.10.0)), I have the following deprecation notices:

> Deprecation Notice: Class Doctrine\Bundle\FixturesBundle\Tests\IntegrationTest\IntegrationTest located in ./vendor/doctrine/doctrine-fixtures-bundle/Tests/IntegrationTest.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///usr/local/bin/composer/src/Composer/Autoload/ClassMapGenerator.php:185
> Deprecation Notice: Class Doctrine\Bundle\FixturesBundle\Tests\IntegrationTest\IntegrationTestKernel located in ./vendor/doctrine/doctrine-fixtures-bundle/Tests/IntegrationTest.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///usr/local/bin/composer/src/Composer/Autoload/ClassMapGenerator.php:185

Is this something you already fixed somewhere?

Thanks

alcaeus:
Fixed in #306, will be included in 3.3.1.

PR #306, opened 2020-03-10
Fix Composer PSR-4 deprecation notices

adrienfr:
This fixes https://github.com/doctrine/DoctrineFixturesBundle/issues/304.


alcaeus:
Thanks @adrienfr!

PR #308, opened 2020-04-02
$container is no longer available in Fixture class

Yondz:
Since ee5eb61, Doctrine\Bundle\FixturesBundle\Fixture does not implement ContainerAwareInterface anymore and thus, $container is not part of this class anymore.

See https://github.com/doctrine/DoctrineFixturesBundle/commit/ee5eb61#diff-8490d6a88059c82df2c98e6f2d0918d2

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -132,10 +132,6 @@ injection::
         $manager->flush();
     }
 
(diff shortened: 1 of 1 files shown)


alcaeus:
LGTM, but should target the current stable branch.

greg0ire:
Thanks @Yondz !

PR #307, opened 2020-03-27
Support for custom purgers

lstrojny:
This PR allows users to specify a custom purger (as previously requested in #116).

As a user, I can register my custom purger factory, that implements `createForEntityManager(?string $emName, EntityManagerInterface $em, array $excluded = [], bool $purgeWithTruncate = false) : PurgerInterface` with the tag `doctrine.fixtures.purger_factory` and some `alias` and then run the command with `--purger=$alias`. If the `ORMPurger` is used `--purge-exclusions`, a list of tables that ought to not be purged, are supported.

PR #312, opened 2020-04-11
Remove deprecation layer

greg0ire:
`doctrine/persistence` already provides a deprecation layer for this by
using class aliases: Doctrine\Common\Persistence\ManagerRegistry and
Doctrine\Persistence\ManagerRegistry are two names for the same thing,
there is no need for another layer here, and the type declaration can be
restored as it was before.


greg0ire:
Sorry @lstrojny , I should have spotted this when reviewing #307, since I'm the author of the deprecation layer in `doctrine/persistence` :sweat_smile:

SenseException (changes requested):
* Resources/doc/index.rst still has `Doctrine\Common\Persistence`
* `Tests/Fixtures/FooBundle` is using `Doctrine\Common\Persistence` as namespace
* I assume `Tests/DeprecationUtil.php` isn't needed anymore, is it?

greg0ire:
Not exactly what this PR aims to fix, but good points! Please review again.

SenseException (approved):

PR #315, opened 2020-04-16
Enable loader to retrieve subtree of fixtures

### LINKED ISSUE #313: How to load subset of registered fixtures?
Hi there,

I'm asking, just in case I missed something by reading the code. I setup fixtures being registered as services, so DI kicks in. This works.

However, within test cases, I want to actually load (as in being put into the storage) a subset of those fixtures. I cannot find a way to do this. Is this currently possible?

havvg:
Fixes #313

### DIFF
--- a/Loader/SymfonyFixturesLoader.php
+++ b/Loader/SymfonyFixturesLoader.php
@@ -13,6 +13,9 @@ use ReflectionClass;
 use RuntimeException;
 use Symfony\Bridge\Doctrine\DataFixtures\ContainerAwareLoader;
 use function array_key_exists;
+use function array_map;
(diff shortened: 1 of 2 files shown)


havvg:
Currently on an upgrade-frenzy, any news on this one?

derrabus:
Seems nobody has picked that up yet. Can you elaborate why you've implemented this feature in the bundle and not in doctrine/data-fixtures?

havvg:
I just implemented it, where the issue was raised. Let me check the library itself.

havvg:
I just implemented it, where the issue was raised. Let me check the library itself.

PR #322, opened 2020-05-29
Allow php 8

greg0ire:
On pause for now, to be continued when Doctrine dependencies allow php 8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,11 @@ jobs:
           php: 7.3
           env: DEPENDENCIES="dev"
 
+        - stage: test
+          php: nightly
+          before_install:
+              - composer config platform.php 7.4.99
+
         # Run phpcs
         - stage: Code Quality
           php: 7.2
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ jobs:
           php: nightly
           before_install:
               - composer config platform.php 7.4.99
+              - composer config minimum-stability dev
 
         # Run phpcs
         - stage: Code Quality
--- a/Tests/Command/LoadDataFixturesDoctrineCommandTest.php
+++ b/Tests/Command/LoadDataFixturesDoctrineCommandTest.php
@@ -10,6 +10,7 @@ use Doctrine\Persistence\ManagerRegistry;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\DependencyInjection\Container;
 use TypeError;
+use const PHP_VERSION_ID;
 
 class LoadDataFixturesDoctrineCommandTest extends TestCase
 {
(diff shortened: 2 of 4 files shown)


esbobkov:
Which dependencies we are waiting?

greg0ire:
Not sure what I meant here… there is a failure though, so let's restart the build in case things have evolved.

greg0ire:
New failures… we have to use PHPUnit 9.3

greg0ire:
Back to square one… the failure that is left is due to error messages being improved in PHP8, and should be easy to fix.

.travis.yml
```diff
@@ -14,6 +14,9 @@ env:
         - SYMFONY_DEPRECATIONS_HELPER="max[self]=0&max[direct]=0"
 
 jobs:
+    allow_failures:
+        - php: nightly
+
```
greg0ire: The feature freeze has started today if we believe https://wiki.php.net/todo/php80, which means I should remove this, right?

Tests/Command/LoadDataFixturesDoctrineCommandTest.php
```diff
@@ -24,6 +25,14 @@ public function testInstantiatingWithoutManagerRegistry() : void
         try {
             new LoadDataFixturesDoctrineCommand($loader);
         } catch (TypeError $e) {
+            if (PHP_VERSION_ID >= 80000) {
+                $this->expectExceptionMessage(
```
stof, line 29: Why do we defined expected exception messages in a `catch` block ? That's weird.

greg0ire: See the `expectedDeprecation` annotation on that method? I think this is why: in the case there is a deprecation, you don't want to expect an exception message at all.

https://github.com/doctrine/DoctrineFixturesBundle/blob/7c4f56d898550a6917b6eb3ae0914b5742dc400f/Tests/Command/LoadDataFixturesDoctrineCommandTest.php#L18-L31

SenseException: Can we keep one solution where only `$this->expectExceptionMessage()` is used instead of the annotation? It would improve readability.

.travis.yml
SenseException: I guess so. But it would be reintroduced later when PHP 8.1 becomes a thing.

Tests/Command/LoadDataFixturesDoctrineCommandTest.php
greg0ire: The annotation is for a deprecation, not for an exception.

SenseException: My bad, I misread it.

.travis.yml
greg0ire: Ok, I dropped it.

SenseException (approved):

Tests/Command/LoadDataFixturesDoctrineCommandTest.php
greg0ire: @stof please review again

ostrolucky (approved):

Issue #320, opened 2020-05-20
Custom Perger is documented on Symfony.com but not included in latest tagged release


thomask:
Documentation [here](https://symfony.com/doc/master/bundles/DoctrineFixturesBundle/index.html) would suggest to me it would be included at least in the latest stable release. However, it is not (same goed for doctrine/data-fixtures).

Thanks!

SenseException:
Neither the 3.1 branch nor the tags contain documentation about the purger. It also seems like the documentation on Symfony is missing some versions.

rcapile:
Any news on this? It seems it was added after the latest release 3.3.1.

xepozz:
Any news??

SenseException:
No news yet, but I lost track of this after my first attempt to communicate this issue with Symfony. Thanks for the reminder.

PR #330, opened 2020-09-01
Allow option to exclude purged tables/views

Zul3s:
Hi, 
I would like add small argument to manage excluded purge tables/views. 
I don't know how i can add test simply ? 

Can be related with #320 ?

### DIFF
--- a/Command/LoadDataFixturesDoctrineCommand.php
+++ b/Command/LoadDataFixturesDoctrineCommand.php
@@ -54,6 +54,7 @@ class LoadDataFixturesDoctrineCommand extends DoctrineCommand
             ->addOption('em', null, InputOption::VALUE_REQUIRED, 'The entity manager to use for this command.')
             ->addOption('shard', null, InputOption::VALUE_REQUIRED, 'The shard connection to use for this command.')
             ->addOption('purge-with-truncate', null, InputOption::VALUE_NONE, 'Purge data by using a database-level TRUNCATE statement')
+            ->addOption('purge-excluded', null, InputOption::VALUE_IS_ARRAY|InputOption::VALUE_OPTIONAL, 'Array of table/view names to be excluded from purge')
             ->setHelp(<<<EOT
 The <info>%command.name%</info> command loads data fixtures from your application:
 
@@ -118,7 +119,8 @@ EOT
 
             return 1;
         }
-        $purger = new ORMPurger($em);
+        $excluded = $input->getOption('purge-excluded');
+        $purger   = new ORMPurger($em, $excluded);
         $purger->setPurgeMode($input->getOption('purge-with-truncate') ? ORMPurger::PURGE_MODE_TRUNCATE : ORMPurger::PURGE_MODE_DELETE);
         $executor = new ORMExecutor($em, $purger);
         $executor->setLogger(static function ($message) use ($ui) : void {
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,11 @@ jobs:
           php: 7.3
           env: DEPENDENCIES="dev"
 
+        - stage: test
+          php: nightly
+          before_install:
+              - composer config platform.php 7.4.99
+
         # Run phpcs
         - stage: Code Quality
           php: 7.2
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ jobs:
           php: nightly
           before_install:
               - composer config platform.php 7.4.99
+              - composer config minimum-stability dev
 
         # Run phpcs
         - stage: Code Quality
--- a/Tests/Command/LoadDataFixturesDoctrineCommandTest.php
+++ b/Tests/Command/LoadDataFixturesDoctrineCommandTest.php
@@ -10,6 +10,7 @@ use Doctrine\Persistence\ManagerRegistry;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\DependencyInjection\Container;
 use TypeError;
+use const PHP_VERSION_ID;
 
 class LoadDataFixturesDoctrineCommandTest extends TestCase
 {
@@ -24,6 +25,14 @@ class LoadDataFixturesDoctrineCommandTest extends TestCase
         try {
             new LoadDataFixturesDoctrineCommand($loader);
         } catch (TypeError $e) {
+            if (PHP_VERSION_ID >= 80000) {
+                $this->expectExceptionMessage(
+                    <<<'MESSAGE'
+Doctrine\Bundle\DoctrineBundle\Command\DoctrineCommand::__construct(): Argument #1 ($doctrine) must be of type Doctrine\Persistence\ManagerRegistry, null given, called in /home/travis/build/doctrine/DoctrineFixturesBundle/Command/LoadDataFixturesDoctrineCommand.php on line 41
+MESSAGE
+                );
+                throw $e;
+            }
             $this->expectExceptionMessage('Argument 1 passed to Doctrine\Bundle\DoctrineBundle\Command\DoctrineCommand::__construct() must be an instance of Doctrine\Persistence\ManagerRegistry, null given');
 
             throw $e;
--- a/Tests/IntegrationTest.php
+++ b/Tests/IntegrationTest.php
@@ -10,6 +10,7 @@ use Doctrine\Bundle\FixturesBundle\Tests\Fixtures\FooBundle\DataFixtures\OtherFi
 use Doctrine\Bundle\FixturesBundle\Tests\Fixtures\FooBundle\DataFixtures\RequiredConstructorArgsFixtures;
 use Doctrine\Bundle\FixturesBundle\Tests\Fixtures\FooBundle\DataFixtures\WithDependenciesFixtures;
 use Doctrine\Common\DataFixtures\Loader;
+use LogicException;
 use PHPUnit\Framework\TestCase;
 use RuntimeException;
 use Symfony\Bridge\Doctrine\DataFixtures\ContainerAwareLoader;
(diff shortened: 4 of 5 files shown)


.idea/DoctrineFixturesBundle.iml
```diff
@@ -0,0 +1,98 @@
...
+      <excludeFolder url="file://$MODULE_DIR$/vendor/symfony/string" />
+      <excludeFolder url="file://$MODULE_DIR$/vendor/symfony/var-dumper" />
+      <excludeFolder url="file://$MODULE_DIR$/vendor/symfony/var-exporter" />
+      <excludeFolder url="file://$MODULE_DIR$/vendor/theseer/tokenizer" />
+      <excludeFolder url="file://$MODULE_DIR$/vendor/webmozart/assert" />
+    </content>
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+  </component>
+</module>
```
greg0ire: You should use `git add -p`, not `git add .`, and then `git commit -v` instead of just `git commit`.

Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -54,6 +54,7 @@ protected function configure()
             ->addOption('em', null, InputOption::VALUE_REQUIRED, 'The entity manager to use for this command.')
             ->addOption('shard', null, InputOption::VALUE_REQUIRED, 'The shard connection to use for this command.')
             ->addOption('purge-with-truncate', null, InputOption::VALUE_NONE, 'Purge data by using a database-level TRUNCATE statement')
+            ->addOption('purge-excluded', null, InputOption::VALUE_IS_ARRAY|InputOption::VALUE_OPTIONAL, 'Table/view names to be excleded from purge')
```
greg0ire: See https://github.com/doctrine/data-fixtures/pull/350 :)

phpunit.xml.dist.bak
```diff
@@ -0,0 +1,24 @@
...
+                <directory>./Tests</directory>
+                <directory>./vendor</directory>
+            </exclude>
+        </whitelist>
+    </filter>
+
+    <listeners>
+        <listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener"/>
+    </listeners>
+</phpunit>
```
greg0ire: Nope

SenseException (changes requested):
Please remove all  `.idea` files from your PR.

.idea/DoctrineFixturesBundle.iml
Zul3s: Yes sorry for that.

phpunit.xml.dist.bak
Zul3s: Yes sorry for that.

Command/LoadDataFixturesDoctrineCommand.php
Zul3s: I am not sur what do you want say ? You think it is useless ?

greg0ire: No, I just think "excluded" should be spelt properly ;)

greg0ire:
This looks like the kind of thing that should be managed with configuration: why would you want to sometimes exclude tables from purge, some other tables not?

Zul3s:
Maybe i don't know the most simple configuration right, but for me, with command parameter it's easy to use. 
example : 
EntityA -> Simple "standard" entity hydrate with fixtures
EntityB -> Entity hydrate by Web API (ex. import european cities) 

On first usage, we want all fixtures and imports but after when we load only fixtures, we don't want delete cities because it's just "reference" data and make long time to import. 

What do you think ?

db306:
> This looks like the kind of thing that should be managed with configuration: why would you want to sometimes exclude tables from purge, some other tables not?

One example are database views mapped to entities, these tables are composed from other tables and purging fails because it tries to truncate them.

johnkary:
My group would really like this PR merged! I have tested it locally and it works great.

> This looks like the kind of thing that should be managed with configuration: why would you want to sometimes exclude tables from purge, some other tables not?

We scaffold our development environment database with a copy of *some* production data such as users, departments, projects, and all their relationships. Our enterprise organization is large and re-creating all these relationships in dev fixtures is too tedious. We purge most production data but keep data from a few tables. And a few tables even select batches of records to keep while purging the rest. Of course this is custom logic we don't expect DoctrineFixturesBundle to support.

We have already created our own Symfony Command that decorates logic around doctrine:fixtures:load, as we recognize "composition over inheritance" when integrating third-party code. Our command first does some custom logic, then programmatically calls doctrine:fixtures:load, then does more custom logic after fixture data exists.

Accepting this PR means our custom command can continue programmatically calling doctrine:fixtures:load like this:

    // This code is inside our custom Symfony Command's execute() method:
    $parameters = [
        "command" => "doctrine:fixtures:load",
        "--append" => $input->getOption('append'),
        "--fixtures" => $ourPathsWithFixtures,
        "--purge-excluded" => [ // NEW LIST GOES HERE
            "table1",
            "table2",
            "table3",
        ],
    ];
    $input = new ArrayInput($parameters);

    $command = $this->getApplication()->find("doctrine:fixtures:load");
    $command->run($input, $output);

There are no other seams available in doctrine:fixtures:load for us to change how purging is done, unless we fork and re-implement doctrine:fixtures:load.

There is no seams to inject custom behavior for purging using ORMPurger, short of hacking up Doctrine metadata. As of today doctrine:fixtures:load does `new ORMPurger($em)`. External code cannot inject its own behavior for that.

The only alternative I can think of is `LoadDataFixturesDoctrineCommand` service accepting a __construct() argument to wire in a custom ORMPurger. I'm betting 99.9% of users would never need a custom ORMPurger, else we'd have an open issue or PR adding it.

    class LoadDataFixturesDoctrineCommand extends DoctrineCommand
    {
        public function __construct(ManagerRegistry $doctrine = null, ORMPurger $purger = null)
        {
            parent::__construct($doctrine);
    
            $this->purger = $purger;
        }
    
        protected function execute(InputInterface $input, OutputInterface $output)
        {
            /** @var $doctrine \Doctrine\Common\Persistence\ManagerRegistry */
            $doctrine = $this->getContainer()->get('doctrine');
            $em = $doctrine->getManager($input->getOption('em'));
    
            // ... snip ...
    
            // NEW LINE HERE
            $purger = $this->purger ?? new ORMPurger($em);
    
            $purger->setPurgeMode($input->getOption('purge-with-truncate') ? ORMPurger::PURGE_MODE_TRUNCATE : ORMPurger::PURGE_MODE_DELETE);

            // ... snip ...
        }
    }

You don't want to hassle with that. Accepting this PR to support custom table purging is much easier for library maintenance and fits most use cases.

@greg0ire Is there anything else stopping this PR from being merged? Tests? Documentation? Please let me know and I'll gladly PR any fixes to implement them and get this merged in.

johnkary (approved):

johnkary:
Ahh, I didn't see that master branch supports a custom purger. We could work with that solution once it's tagged in a stable release! Thanks for all your hard work.

db306:
@johnkary while this gets released my approach was to have a custom purge cli method of my own that purges the databases followed by a `doctrine:fixtures:load --append`. 

By doing so you can purge exactly what and how you want it and just load the fixtures natively. 

In the example below I want to drop all tables and delete views from my postgresql db.

Hope this can help on the mean time.

```php
use Doctrine\ORM\EntityManagerInterface;
use Doctrine\ORM\Query\ResultSetMapping;
use Symfony\Component\Console\Command\Command;
use Symfony\Component\Console\Input\InputInterface;
use Symfony\Component\Console\Output\OutputInterface;

class DropTestTables extends Command
{
    private string $env;
    private EntityManagerInterface $entityManager;
    protected static $defaultName = 'test:drop';

    public function __construct(
        string $env,
        EntityManagerInterface $entityManager
    ) {
        $this->env = $env;
        $this->entityManager = $entityManager;
        parent::__construct();
    }

    protected function configure()
    {
        $this
            ->setDescription('Deletes all tables and sets everything to 0')
            ->setHelp(
                'This command allows the system to auto-purge its db structure in order to be useable by functional tests. As of today, no other method exists in order to empty dbs and hanlding views'
            );
    }

    protected function execute(InputInterface $input, OutputInterface $output)
    {
        if ('test' !== $this->env) {
            $output->writeln('❌  YOU ARE ONLY ALLOWED TO RUN THIS IN TEST ENVIRONMENT');

            return Command::FAILURE;
        }

        $output->writeln(
            [
                'DROP ALL TABLES IN TEST ENVIRONMENT ',
                '=====================================',
            ]
        );

        $sm = $this->entityManager->getConnection()->getSchemaManager();
        $viewsRaw = array_values($sm->listViews());
        $views = [];
        $tables = array_values($sm->listTables());
        $dropTables = 'DROP TABLE IF EXISTS ';
        $dropViews = 'DROP VIEW IF EXISTS ';

        foreach ($viewsRaw as $i => $view) {
            if (str_starts_with($view->getName(), 'public.')) {
                $viewName = str_replace('public.', '', $view->getName());
                $dropViews .= count($views) ? ", ${viewName}" : $viewName;
                $views[] = $viewName;
            }
        }

        foreach ($tables as $i => $table) {
            $tableName = $table->getName();
            $dropTables .= $i ? ", ${tableName}" : $tableName;
        }

        if (count($views)) {
            $this->entityManager->createNativeQuery($dropViews, new ResultSetMapping())->execute();
        }

        $output->writeln('✅ Dropped '.count($views).' Views from db');

        if (count($tables)) {
            $this->entityManager->createNativeQuery($dropTables, new ResultSetMapping())->execute();
        }

        $output->writeln('✅ Dropped '.count($tables).' Tables from db');

        return Command::SUCCESS;
    }
}
```

greg0ire:
This PR should target `master` since it is not a bugfix. I'm going to rename `master` to `3.4.x` at some point, so that it's less confusing for everyone. Other than than, I think the PR looks good, and my question has been answered by several people with different use cases, showing there is indeed a need for this.

greg0ire (approved):

greg0ire:
Apparently there are conflicts, maybe with the custom purger you were talking about. I have some maintenance work to do here, I'll do it and then I will release 3.4.0

johnkary:
Looks like this PR is no longer needed because LoadDataFixturesDoctrineCommand in master supports option `--purge-exclusions` which has the same behavior as this PR. See https://github.com/doctrine/DoctrineFixturesBundle/blame/master/Command/LoadDataFixturesDoctrineCommand.php#L65

greg0ire:
Let's close then. I'm working on fixing the build at #334

greg0ire:
I released. https://twitter.com/greg0ire/status/1327553483118096386

PR #338, opened 2021-01-12
Fixed non-interactive mode of doctrine:fixtures:load

rapliandras:
The command was previously interrupting in non-interactive mode.


rapliandras:
The CI issue is not caused by the patch!

SenseException:
> The CI issue is not caused by the patch!

Seems like it. Despite that, your changes need to be covered by tests.

greg0ire:
I fixed the CI issue, so all is left to do is write the tests :)

> The command was previously interrupting in non-interactive mode.

What do you mean? Can you give more details, and explain why/how your patch fixes the issue? It's not obvious to me.

derrabus:
No response from the contributor. Let's close.

PR #340, opened 2021-01-24
Removing internal link

ThomasLandauer:
I guess this is the first time that I actually want to *remove* a link :-) But this one leads just one (!) screen down to a self-explanatory heading with the exact same wording. So I would say, it just adds noise.

I you disagree, I'll add the word "below" to signify that it's in fact just a jump link.

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -74,10 +74,6 @@ Create a fixtures class and start adding products::
         }
     }
 
-.. tip::
-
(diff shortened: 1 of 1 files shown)


greg0ire (approved):

SenseException (approved):

SenseException:
I think that's okay. The docs of DoctrineFixturesBundle are written in steps which is why it's not advisable to skip a part.

greg0ire:
Thanks @ThomasLandauer !

PR #343, opened 2021-02-14
Bump symfony/dependency-injection

greg0ire:
The --prefer-lowest job fails with an error message that is different
from the 7.1 job, and that seems to be related to dependency injection.


greg0ire:
Wow, I did not expect that to work the first time!

PR #344, opened 2021-02-14
Remove workaround for testing with PHP 8

greg0ire:
Now that all dependencies support it, it should not be necessary to
pretend we are using PHP 7.4. Implies upgrading the CS lib.

### DIFF
--- a/Tests/Command/LoadDataFixturesDoctrineCommandTest.php
+++ b/Tests/Command/LoadDataFixturesDoctrineCommandTest.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\Bundle\FixturesBundle\Tests\Command;
 
+use Doctrine\Bundle\DoctrineBundle\Command\DoctrineCommand;
 use Doctrine\Bundle\FixturesBundle\Command\LoadDataFixturesDoctrineCommand;
 use Doctrine\Bundle\FixturesBundle\Loader\SymfonyFixturesLoader;
 use Doctrine\Persistence\ManagerRegistry;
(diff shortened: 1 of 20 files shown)


Tests/Command/LoadDataFixturesDoctrineCommandTest.php
```diff
@@ -29,11 +31,13 @@ public function testInstantiatingWithoutManagerRegistry() : void
             if (PHP_VERSION_ID >= 80000) {
                 $this->expectExceptionMessage(
                     <<<'MESSAGE'
-Doctrine\Bundle\DoctrineBundle\Command\DoctrineCommand::__construct(): Argument #1 ($doctrine) must be of type Doctrine\Persistence\ManagerRegistry, null given, called in /home/runner/work/DoctrineFixturesBundle/DoctrineFixturesBundle/Command/LoadDataFixturesDoctrineCommand.php on line 49
+Doctrine\Bundle\DoctrineBundle\Command\DoctrineCommand::__construct(): Argument #1 ($doctrine) must be of type Doctrine\Persistence\ManagerRegistry, null given, called in /home/runner/work/DoctrineFixturesBundle/DoctrineFixturesBundle/Command/LoadDataFixturesDoctrineCommand.php on line 51
```
SenseException: Is the line number relevant for the error message check? It seems like an unstable test when changes are happening.

SenseException (approved):

Tests/Command/LoadDataFixturesDoctrineCommandTest.php
greg0ire: It is unstable indeed, I remember having to change it in the past. Let me try refactoring it.

greg0ire:
It looks like the integration tests are flaky :facepalm: 

```
 There was 1 error:

1) Doctrine\Bundle\FixturesBundle\Tests\IntegrationTest::testRunCommandWithPurgeMode
Symfony\Component\DependencyInjection\Exception\InvalidArgumentException: The "doctrine" service is private, you cannot replace it.

/home/runner/work/DoctrineFixturesBundle/DoctrineFixturesBundle/vendor/symfony/dependency-injection/Container.php:169
/home/runner/work/DoctrineFixturesBundle/DoctrineFixturesBundle/Tests/IntegrationTest.php:483
```

PR #347, opened 2021-03-22
Add metadata file for symfony.com

greg0ire:
@javiereguiluz please review


.symfony.bundle.yaml
```diff
@@ -0,0 +1,4 @@
+branches: ["3.3.x", "3.4.x"]
```
javiereguiluz: `3.5.x` should be added both inside `branches` and `maintained_branches`

javiereguiluz (commented):
Thanks! This looks correct to me except for the minor comment left in the review.

PR #355, opened 2021-10-17
Reuse workflows and setup static analysis

greg0ire:


.github/workflows/coding-standards.yml
```diff
@@ -12,29 +12,4 @@ on:
...
-
-      - name: "Install dependencies with Composer"
-        uses: "ramsey/composer-install@v1"
-        with:
-          dependency-versions: "highest"
-
-      # https://github.com/doctrine/.github/issues/3
-      - name: "Run PHP_CodeSniffer"
-        run: "vendor/bin/phpcs -q --no-colors --report=checkstyle | cs2pr"
+    uses: "doctrine/.github/.github/workflows/coding-standards.yml@1.1.1"
```
SenseException, line 15: Was the switch to PHP 8.0 on purpose?

greg0ire: Yes, it does not matter here IMO.

SenseException (approved):

PR #360, opened 2021-11-17
Update index.rst

BahmanMehrdad:
Added separate code-block for blocks.

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -152,7 +152,7 @@ Sharing Objects between Fixtures
 When using multiple fixtures files, you can reuse PHP objects across different
 files thanks to the **object references**. Use the ``addReference()`` method to
 give a name to any object and then, use the ``getReference()`` method to get the
-exact same object via its name::
+exact same object via its name:
 
 .. code-block:: php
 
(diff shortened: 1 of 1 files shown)


Resources/doc/index.rst
```diff
@@ -154,6 +154,8 @@ files thanks to the **object references**. Use the ``addReference()`` method to
 give a name to any object and then, use the ``getReference()`` method to get the
 exact same object via its name::
 
+.. code-block:: php
```
greg0ire, line 157: Is it syntactically valid to have `::`, then this? I don't like what I see when I click "rich diff"

BahmanMehrdad: This is correct in terms of grammatical structure.
See this link:
https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block

SenseException (approved):

Resources/doc/index.rst
greg0ire: I'm not talking about the `::` on line 157, but about the one on line 155.

BahmanMehrdad: "A paragraph consisting of two colons ("::") signifies that the following text block(s) comprise a literal block. The literal block must either be indented or quoted (see below). No markup processing is done within a literal block. It is left as-is, and is typically rendered in a monospaced typeface"

But here we do not need it because the two blocks are defined.
The code was edited.

greg0ire: And now the rich diff looks good. Thanks.

BahmanMehrdad: @greg0ire NP ;)

PR #362, opened 2021-12-06
UserPasswordEncoderInterface depricated

Maxime-p:
Replacing UserPasswordEncoderInterface (depricated since Symfony 5.3) by UserPasswordHasherInterface

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -108,15 +108,15 @@ class. No problem! Your fixtures class is a service, so you can use normal depen
 injection::
 
     // src/DataFixtures/AppFixtures.php
-    use Symfony\Component\Security\Core\Encoder\UserPasswordEncoderInterface;
+    use Symfony\Component\PasswordHasher\Hasher\UserPasswordHasherInterface;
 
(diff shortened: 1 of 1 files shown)


derrabus (approved):

SenseException (approved):
Thanks for your first PR in this repository. All in all this is a valid change and approvable, but the used `UserPasswordHasherInterface` interface and the component it belongs to is something that was introduced in Symfony 5.3, while this bundle still supports Symfony >3.4. It won't work for too many people.

Maxime-p:
> Thanks for your first PR in this repository. All in all this is a valid change and approvable, but the used `UserPasswordHasherInterface` interface and the component it belongs to is something that was introduced in Symfony 5.3, while this bundle still supports Symfony >3.4. It won't work for too many people.

Ok, so can we do something else than waiting? Sorry I'm new in the contributing world ^^

greg0ire:
IMO For docs, with should assume Symfony 6, DBAL 3, and PHP 8.1 or whatever the latest versions are when we write the docs. Let's be nice to people who did the homework instead of giving them more deprecations to fix.

PR #365, opened 2022-04-28
Allow doctrine/persistence 3

hlecorche:


greg0ire:
The build fails: 

> Calling `transactional()` instead of `wrapInTransaction()` which is not implemented on Mock_EntityManagerInterface_69c8e1e4 

:thinking:

hlecorche:
@greg0ire That's right. 🙁
The problem is not related to this PR (the problem was already present before). I will open a new ticket.

derrabus (approved):

PR #375, opened 2023-01-04
chore: add ci tests for php 8.2

Chris53897:
bump github action version for checkout 2 => 3

The failing test is not related to php 8.2.
If i run tests locally with php 8.2 (and php 8.1) i get different results after each run.
range is from 0 to 3 errors.

`Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException : You have requested a non-existent service "test.doctrine.fixtures.loader". Did you mean this: "test.doctrine.fixtures_load_command"?`

### DIFF
--- a/Tests/IntegrationTestKernel.php
+++ b/Tests/IntegrationTestKernel.php
@@ -25,7 +25,7 @@ class IntegrationTestKernel extends Kernel
 
     public function __construct(string $environment, bool $debug)
     {
-        $this->randomKey = rand(100, 999);
+        $this->randomKey = rand(10000000000, 99999999999);
 
         parent::__construct($environment, $debug);
     }
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -29,6 +29,7 @@ jobs:
           - "7.4"
           - "8.0"
           - "8.1"
+          - "8.2"
         dependencies:
           - "highest"
         stability:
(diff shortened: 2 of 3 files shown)


greg0ire:
Indeed if I re-run the tests, fewer are failing… I don't know why that is.

greg0ire:
The remaining failure seems similar to https://github.com/doctrine/data-fixtures/issues/414

Tests/IntegrationTestKernel.php
```diff
@@ -25,7 +25,7 @@ class IntegrationTestKernel extends Kernel
 
     public function __construct(string $environment, bool $debug)
     {
-        $this->randomKey = rand(100, 999);
```
Chris53897, line 28: This does fix the local runs for me. I got zero errors with Php 8.1 and 8.2 local.

Chris53897: This view in Github is misleading. I changed it to `$this->randomKey = rand(10000000000, 99999999999);`
Not a real solution, but good enough for the moment i guess.

greg0ire: How is it misleading? I understood that was what you did. Nice find, it would make sense that this is responsible for random failures.

Chris53897:
The remaining PhpStan error should be resolved in a sperate PR in my eyes. WDYT?

greg0ire:
I agree.

Tests/IntegrationTestKernel.php
Chris53897: Github only shows in the preview that the line was removed, not that it is replaced.

Tests/IntegrationTestKernel.php
```diff
@@ -25,7 +25,7 @@ class IntegrationTestKernel extends Kernel
...
+        $this->randomKey = rand(10000000000, 99999999999);
```
greg0ire, line 28: > Github only shows in the preview that the line was removed, not that it is replaced.


It does show it here

greg0ire (approved):

greg0ire:
Thanks @Chris53897 !

PR #382, opened 2023-04-11
Add `return void` annotation to silence deprecation msg

PhilETaylor:
> can you please send a PR to add the `@return void` annotation? That's enough to express the intent and fix the deprecation.

ref https://github.com/doctrine/DoctrineFixturesBundle/pull/381#issuecomment-1502825960

Used single/one line annotation syntax as [proposed elsewhere](https://github.com/doctrine/DoctrineMigrationsBundle/pull/492#discussion_r1150441505) 

//cc @nicolas-grekas


derrabus (approved):

derrabus:
Thank you!

PR #389, opened 2023-05-15
Add code example for purge exclusion

LucileDT:
There is a section talking about excluding tables from the fixtures purge but no example, making it hard to see this feature exists. This commit adds an example for this command option.

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -302,7 +302,11 @@ By default all previously existing data is purged using ``DELETE FROM table`` st
 
 If you want to exclude a set of tables from being purged, e.g. because your schema comes with pre-populated,
 semi-static data, pass the option ``--purge-exclusions``. Specify ``--purge-exclusions`` multiple times to exclude
-multiple tables.
+multiple tables:
+
+.. code-block:: terminal
+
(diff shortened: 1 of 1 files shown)


Resources/doc/index.rst
```diff
@@ -302,7 +302,11 @@ By default all previously existing data is purged using ``DELETE FROM table`` st
 
 If you want to exclude a set of tables from being purged, e.g. because your schema comes with pre-populated,
 semi-static data, pass the option ``--purge-exclusions``. Specify ``--purge-exclusions`` multiple times to exclude
-multiple tables.
+multiple tables:
+
+.. code-block:: terminal
+
+    $ php bin/console doctrine:fixtures:load --exclude-table=post_category --exclude-table=comment_type
```
greg0ire: :thinking: shouldn't it be `--purge-exclusions`?

```suggestion
    $ php bin/console doctrine:fixtures:load --purge-exclusions=post_category --purge-exclusions=comment_type
```

LucileDT: Oh yes you're right, my bad

greg0ire:
Thanks @LucileDT !

PR #391, opened 2023-05-23
Add a note that object references only work for Entities

mpdude:
At least, that's what it looks like from the code.


greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @mpdude!

PR #398, opened 2023-10-23
Adding int return type in command

javiereguiluz:
Seen while trying to upgrade a Symfony app to Symfony 7.0.

```
PHP Fatal error: 
Declaration of Doctrine\Bundle\FixturesBundle\Command\LoadDataFixturesDoctrineCommand::execute
(InputInterface $input, OutputInterface $output)
must be compatible with Symfony\Component\Console\Command\Command::execute
(InputInterface $input, OutputInterface $output): int
in vendor/doctrine/doctrine-fixtures-bundle/Command/LoadDataFixturesDoctrineCommand.php on line 99
```

### DIFF
--- a/Command/LoadDataFixturesDoctrineCommand.php
+++ b/Command/LoadDataFixturesDoctrineCommand.php
@@ -93,10 +93,7 @@ EOT
         );
     }
 
-    /**
-     * @return int
-     */
-    protected function execute(InputInterface $input, OutputInterface $output)
+    protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $ui = new SymfonyStyle($input, $output);
 


derrabus:
Thank you. We probably need a solution for lower branches as well because we don't have plans to release v4 just yet. A conditional trait like I used in DBAL and ORM might do the trick there.

javiereguiluz:
Thanks for the insights. I guess you refer to doing something like this: https://github.com/doctrine/dbal/pull/6136/files

I don't know the Doctrine project code well enough, so I'm closing this PR and I will let you (or any other volunteer contributor) to do the proper fix based on the trait. Thanks a lot!

derrabus:
Exactly, that's the PR I'm referring to.

PR #403, opened 2023-10-28
Drop support for Symfony 3 + 4

derrabus:
This PR drops the support for PHP 7.1 - 7.3 and Symfony 3 - 5.3. We've done the same on the data fixtures repo with doctrine/data-fixtures#438 and I believe it only makes sense to keep the requirements of the bundle in sync with the library it's supposed to integrate.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -23,9 +23,6 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - "7.1"
-          - "7.2"
-          - "7.3"
           - "7.4"
           - "8.0"
           - "8.1"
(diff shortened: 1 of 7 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -23,28 +23,18 @@ jobs:
...
-          - dependencies: "highest"
-            extra_constraint: "symfony/lts:v3"
-            php-version: "7.3"
-          - dependencies: "highest"
-            stability: "dev"
-            php-version: "7.3"
+            php-version: "7.4"
           - dependencies: "highest"
             stability: "dev"
             php-version: "8.2"
```
SenseException: Isn't this supposed to be 8.3? Or maybe a

```suggestion
            php-version: "8.2"
          - dependencies: "highest"
            stability: "dev"
            php-version: "8.3"
```

is needed?

derrabus: Yes. I've done this in #405 now.

SenseException (approved):

PR #404, opened 2023-10-28
Doctrine CS 12

derrabus:


composer.json
```diff
@@ -32,7 +32,7 @@
         "symfony/http-kernel": "^3.4|^4.3|^5.0|^6.0"
     },
     "require-dev": {
-        "doctrine/coding-standard": "^9",
+        "doctrine/coding-standard": "^9 || ^12",
```
SenseException, line 35: This 9.0.1 and 9.0.2 thing with PHP 7.1 still gets me. :grinning:

SenseException (approved):

PR #421, opened 2024-01-16
Automatically extract fixture dependencies

### LINKED ISSUE #371: Possibility to add fixtures into dependent group
I have strange behaviour when attempting to load fixture group by class name with dependent fixtures.

Example:
Let's create two fixture classes:
```php
<?php

declare(strict_types=1);

namespace App\DataFixtures;

use Doctrine\Bundle\FixturesBundle\Fixture;
use Doctrine\Persistence\ObjectManager;

class InitFixtures extends Fixture
{
    public function load(ObjectManager $manager): void
    {
        // load init fixtures
    }
}
```
```php
<?php

declare(strict_types=1);

namespace App\DataFixtures;

use Doctrine\Bundle\FixturesBundle\Fixture;
use Doctrine\Persistence\ObjectManager;

class DevFixtures extends Fixture implements DependentFixtureInterface
{
    public function load(ObjectManager $manager): void
    {
        // load dev fixtures
    }

    public function getDependencies(): array
    {
        return [
            InitFixtures::class,
        ];
    }
}
```

Then runing `php bin/console doctrine:fixtures:load --group=DevFixtures --no-interaction` caused error
```
In SymfonyFixturesLoader.php line 148:

  Fixture "App\DataFixtures\InitFixtures" was declared as a dependency for fixture "App\DataFixtures\DevFixtures", but it was not included in any of the loaded fixture groups.
```
Because InitFixtures doesn't have such group.

It would be great if dependent fixtures will be added automatically into groups, which from they depend. Like it is made in main library - https://github.com/doctrine/data-fixtures/blob/1.5.x/lib/Doctrine/Common/DataFixtures/Loader.php#L171-L177

cs278:
Fixes https://github.com/doctrine/DoctrineFixturesBundle/issues/371

This improves the cooperation between the native fixture dependencies and the bundle provided groups. It is now no longer necessary to add the dependencies of fixtures in a group into the group as well.

I've not tested this extensively, or added tests for this yet but if the maintainers are happy with the change in behaviour I can do that.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -5,6 +5,11 @@ parameters:
 
 	ignoreErrors:
 		-
+			message: "#^Call to an undefined method Doctrine\\\\Bundle\\\\FixturesBundle\\\\Loader\\\\SymfonyFixturesLoader\\:\\:getFixture\\(\\)\\.$#"
+			count: 1
+			path: src/Loader/SymfonyFixturesLoader.php
+
+		-
 			message: "#^Call to an undefined static method Doctrine\\\\Bundle\\\\FixturesBundle\\\\Loader\\\\SymfonyBridgeLoader\\:\\:addFixture\\(\\)\\.$#"
 			count: 1
 			path: src/Loader/SymfonyFixturesLoader.php
--- a/src/Loader/SymfonyFixturesLoader.php
+++ b/src/Loader/SymfonyFixturesLoader.php
@@ -10,9 +10,8 @@ use Doctrine\Common\DataFixtures\DependentFixtureInterface;
 use Doctrine\Common\DataFixtures\FixtureInterface;
 use LogicException;
 use ReflectionClass;
-use RuntimeException;
 
-use function array_key_exists;
+use function array_keys;
 use function array_values;
 use function get_class;
 use function sprintf;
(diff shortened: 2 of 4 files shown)


cs278:
I'd appreciate some feedback on the behaviour change before I do any more work making this suitable to be merged.

greg0ire:
I don't know much about this bundle, but I don't see a problem with this, please go ahead.

cs278:
> I don't know much about this bundle, but I don't see a problem with this, please go ahead.

@greg0ire :+1: updated

greg0ire:

Loader/SymfonyFixturesLoader.php
```diff
@@ -130,22 +132,25 @@ private function addGroupsFixtureMapping(string $className, array $groups): void
     }
 
     /**
-     * @param string[] $fixtures An array of fixtures with class names as keys
+     * Collect any dependant fixtures from the given classes.
```
greg0ire:
```suggestion
     * Collect any dependent fixtures from the given classes.
```

cs278:
> It seems there are CI jobs failing. Please take a look at [this guide](https://www.doctrine-project.org/contribute/#dealing-with-checks-and-tools) for more on how to handle those.

PHPStan is failing but it seems to be it doesn't understand what the parent class is so it thinks the methods are undefined, it's doing the same on the `3.5.x` branch so nothing I've done is causing it.

greg0ire:
Are you sure? Here is a 12 hours old build not exhibiting any issue: https://github.com/doctrine/DoctrineFixturesBundle/actions/runs/8513966759

greg0ire:
Here is another PR I just made: https://github.com/doctrine/DoctrineFixturesBundle/pull/429
It's green.

cs278:
Apologies, I merged from the wrong remote which meant I didn't have b83a046 I've ignored the PHPStan error I've introduced which is the same class as the existing ignored problems.

greg0ire:
Retargeting on 3.6.x since this is no bugfix. Please rebase and force push.

cs278:
@greg0ire updated

greg0ire:
I don't think it makes sense to have 2 commits here: we're never going to want to revert one and keep the other, right? If so, please squash them, preserving the excellent commit message body.

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @cs278 !

cs278:
Thanks for your assistance in getting this merged @greg0ire :smiley:

PR #426, opened 2024-03-29
Switch to a modern directory structure

greg0ire:
Here are the steps I followed:

- `mkdir src`
- `git mv *.php src`
- `git mv Command/ DependencyInjection Loader Purger src`
- adjust autoloading path
- overrode `Bunde::getPath()`
- `git mv Resources/doc/ docs`
- `git mv Resources/config/ config`
- `git mv Tests/ tests`


composer.json
```diff
@@ -43,7 +43,10 @@
         "doctrine/dbal": "< 3"
     },
     "autoload": {
-        "psr-4": { "Doctrine\\Bundle\\FixturesBundle\\": "" }
+        "psr-4": { "Doctrine\\Bundle\\FixturesBundle\\": "src" }
+    },
+    "autoload-dev": {
+        "psr-4": { "Doctrine\\Bundle\\FixturesBundle\\Tests\\": "tests" }
```
greg0ire, line 49: This is technically a breaking change, because `Tests` is no longer autoloadable by users of this bundle. Looking at the files in there though, there does not seem to be any code that should be.

SenseException: The tests-directory shouldn't even be considered as usable outside of the bundle itself.

SenseException (approved):

PR #427, opened 2024-03-29
Configure docs path depending on the branch

greg0ire:
The directory structure differs between both branches.


SenseException (approved):

SenseException:
Does the 3.6.x branch still need the changes of the directory structure that is referred in the change?

greg0ire:
@SenseException sorry, I'm not sure what you mean. We can merge this up and simplify it. My understanding is that only the file in the default branch matters.

SenseException:
Never mind, it got covered with one of the other PRs that are now merged.

PR #434, opened 2024-05-02
Address deprecation of HttpKernel Extension class

greg0ire:
The right class to use now is in the DI component, which does make a lot more sense.


SenseException (approved):

javiereguiluz:
Thanks for fixing this deprecation. When you can, please tag a new 3.6.x release so we can get rid of this deprecation when upgrading apps to Symfony 7.1. Thanks!

greg0ire:
@javiereguiluz done :slightly_smiling_face:

javiereguiluz:
Thank you!

PR #441, opened 2024-05-07
Update Symfony bundle metadata

javiereguiluz:
The building of the docs of 3.5.x branch started failing on symfony.com. We need to update this config because 3.5.x stores the docs in a different path.

### DIFF
--- a/.symfony.bundle.yaml
+++ b/.symfony.bundle.yaml
@@ -8,5 +8,5 @@ maintained_branches:
     - "3.5.x"
     - "3.6.x"
     - "3.7.x"
-doc_dir: "docs/"
+doc_dir: {"3.5.x": "Resources/doc/", "3.6.x": "docs/", "3.7.x": "docs/"}
 dev_branch: "3.7.x"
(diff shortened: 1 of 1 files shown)


.symfony.bundle.yaml
```diff
@@ -8,5 +8,5 @@ maintained_branches:
     - "3.5.x"
```
greg0ire: Sorry, I caused this issue. The right thing to do is to remove `3.5.x` from `maintained_branches` I think.

```suggestion
```

.symfony.bundle.yaml
```diff
@@ -8,5 +8,5 @@ maintained_branches:
...
     - "3.6.x"
     - "3.7.x"
-doc_dir: "docs/"
+doc_dir: {"3.5.x": "Resources/doc/", "3.6.x": "docs/", "3.7.x": "docs/"}
```
greg0ire:
```suggestion
doc_dir: "docs/"
```

javiereguiluz: OK, I added another commit to remove `3.5.x` from maintained versions. Thanks.

fabpot (approved):

.symfony.bundle.yaml
greg0ire: Thank you!

Issue #420, opened 2023-12-18
Since symfony/dependency-injection 6.4: "Symfony\Bridge\Doctrine\DataFixtures\ContainerAwareLoader" is deprecated, use dependency injection in your fixtures instead. 


ruudk:
How can we silence this deprecation? We're not using the `ContainerAwareLoader`.

We're using:
```
symfony/doctrine-bridge               v6.4.0
doctrine/data-fixtures                1.7.0
doctrine/doctrine-fixtures-bundle     3.5.1
```

I saw this problem being predicted in https://github.com/doctrine/DoctrineFixturesBundle/pull/395 but I don't think there is a solution yet.

derrabus:
The warning should be gone if you upgrade the bridge to v7, but I realize that this might not be the answer you're looking for. 🙈 

Maybe we could add an option to opt-out of using that class?

BackNot:
I guess we need to wait for the 4.0 version

ruudk:
Upgrading to 7.0 solves it for me.

derrabus:
Thanks for the feedback, but let's leave the issue open. We should try to find a solution for Symfony 6.4 projects.

stanislav-oop:
Is this fixed yet?

derrabus:
No.

Natizyskunk:
Still no solution ?
I would prefer not to mix package versions in my composer.json sinc all my project dependencies are in 6.4.

greg0ire:
And I would prefer if you folks explored the issue and proposed solutions instead of demanding updates and filling my inbox.

stanislav-oop:
Due to some internal company policies we can not update to anything but latest symfony LTS version which is 6.4.
Nothing of proposed solutions works for us so we would like to have this depreciation silenced. Its not a demand though.

derrabus:
Folks.

* Nobody says that you cannot use Symfony 6.4 anymore.
* Nobody's actively working on a solution yet.
* Anyone can open a PR and propose a solution.

I hope that answers all open questions. :v:

sofigio:
Any way to have this feature on Ubuntu 24, without waiting for the next LTS release in 2026? Like porting it to the 1.4.x... or is there any APT repository available with 1.5?

PR #446, opened 2024-07-14
Fix deprecations and ensure compatibility with Symfony 6.4 (#420) for…

joscraw:
Fix deprecations and ensure compatibility with Symfony 6.4 (#420)

- Updated `SymfonyFixturesLoader` to conditionally set the container:
  - Added a `class_exists` check for `ContainerAwareInterface` in the `addFixture` method.
  - Passed `bool` and `ContainerInterface` to the constructor of `SymfonyFixturesLoader`.

- Updated `DoctrineFixturesExtension` for future compatibility:
  - Added `?ConfigurationInterface` return type to `getConfiguration` method.
  - Replaced `HttpKernelExtension` with `Symfony\Component\DependencyInjection\Extension\Extension`.

- Fixed tests to correctly initialize `SymfonyFixturesLoader`:
  - Updated `LoadDataFixturesDoctrineCommandTest` to pass correct parameters to `SymfonyFixturesLoader`.

- Implemented `isUninitializedObject` method in `EntityManagerDecorator` for future Doctrine compatibility.

These changes address deprecation warnings and ensure compatibility with Symfony 6.4 and future versions.

### DIFF
--- a/src/DependencyInjection/DoctrineFixturesExtension.php
+++ b/src/DependencyInjection/DoctrineFixturesExtension.php
@@ -10,8 +10,8 @@ use Exception;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
-use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\DependencyInjection\Extension\Extension;
+use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 
 use function dirname;
 
@@ -21,6 +21,7 @@ class DoctrineFixturesExtension extends Extension
      * {@inheritDoc}
      *
      * @return void
+     *
      * @throws Exception
      */
     public function load(array $configs, ContainerBuilder $container): void
@@ -30,7 +31,7 @@ class DoctrineFixturesExtension extends Extension
         $loader->load('services.xml');
 
         $configuration = $this->getConfiguration($configs, $container);
-        $config = $this->processConfiguration($configuration, $configs);
+        $config        = $this->processConfiguration($configuration, $configs);
 
         $container->setParameter('doctrine_fixtures.use_container_aware_loader', $config['use_container_aware_loader']);
 
@@ -38,6 +39,13 @@ class DoctrineFixturesExtension extends Extension
             ->addTag(FixturesCompilerPass::FIXTURE_TAG);
     }
 
+    /**
+     * {@inheritDoc}
+     *
+     * @param array $config
+     *
+     * @return ConfigurationInterface|null
+     */
     public function getConfiguration(array $config, ContainerBuilder $container): ?ConfigurationInterface
     {
         return new Configuration();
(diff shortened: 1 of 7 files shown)


craigh:
bump 🤜

greg0ire:

craigh:
@joscraw any chance you can update this to fix the failing jobs?

joscraw:
@craigh Hey guys! Sorry, I was out for a bit. Yup I'm on it! I'll take care of it here in a few min.

stanislav-oop:
Thank you! When do you think it will be merged?

src/DependencyInjection/DoctrineFixturesExtension.php
```diff
@@ -19,14 +21,33 @@ class DoctrineFixturesExtension extends Extension
      * {@inheritDoc}
      *
      * @return void
+     *
+     * @throws Exception
```
derrabus, line 25: This annotation is pointless. Any method can throw arbitrary exceptions.

src/DependencyInjection/DoctrineFixturesExtension.php
```diff
@@ -19,14 +21,33 @@ class DoctrineFixturesExtension extends Extension
...
      */
-    public function load(array $configs, ContainerBuilder $container)
+    public function load(array $configs, ContainerBuilder $container): void
```
derrabus, line 27: This is a breaking change. We will include native return types in the 4.0.x release, but we cannot do this in a minor release.

src/DependencyInjection/DoctrineFixturesExtension.php
```diff
@@ -19,14 +21,33 @@ class DoctrineFixturesExtension extends Extension
...
     {
         $loader = new XmlFileLoader($container, new FileLocator(dirname(__DIR__) . '/../config'));
 
         $loader->load('services.xml');
 
+        $configuration = $this->getConfiguration($configs, $container);
+        $config        = $this->processConfiguration($configuration, $configs);
+
+        $container->setParameter('doctrine_fixtures.use_container_aware_loader', $config['use_container_aware_loader']);
```
derrabus, line 36: You're building a feature flag for a deprecated feature. That would only make sense if we instantly deprecated not opting out of the container-aware loader.

src/Loader/SymfonyBridgeLoader.php
```diff
@@ -5,18 +5,8 @@
...
-    {
-    }
-} else {
-    /** @internal */
-    abstract class SymfonyBridgeLoader extends Loader
-    {
-    }
+/** @internal */
+abstract class SymfonyBridgeLoader extends Loader
+{
```
derrabus, line 11: The internal `SymfonyBridgeLoader` class was introduced because we needed to conditionally extend the `ContainerAwareLoader` in order to not break BC. Leaving this as an empty class is pretty pointless.

This class has been removed on the 4.0.x branch already.

derrabus (changes requested):
Sorry that it took us so long to pick this PR up. I don't think it makes sense to merge it, tbh. If not using the container-aware stuff at all is the only way to get rid of the deprecations, we should rather release the 4.0.x branch instead.

stanislav-oop:
I am reviewing the depreciations (LTS symfony 6.4) and I can see the depreciation is coming from just the doctrine fixtures without any other custom code. Can not find any other workaround.

derrabus:
> Can not find any other workaround.

Does that fix the deprecations for you?

```JSON
"doctrine/doctrine-fixtures-bundle": "^4@dev"
```

stanislav-oop:
Yes it does, but for the minimum-stablity

derrabus:
> the minimum-stablity

That's fixable on our side. 🙉

@greg0ire @SenseException Shall we release 4.0?

greg0ire:
[Sure!](https://github.com/doctrine/DoctrineFixturesBundle/releases/tag/4.0.0)

joscraw:
@derrabus Releasing the 4.0.x branch instead def makes more sense to me! Thanks guys!

PR #450, opened 2024-10-21
Update index.rst for service detection.

DocFX:
Added precision on service detection.

Took me a couple hours to debug this one, was sweating a lot. ^^
Hope it's clear enough to better help people.

Edit: should we mention to use `php bin/console debug:container --tag doctrine.fixture.orm` to check the services?

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -77,7 +77,7 @@ using the `default service configuration`_, any class that implements ``ORMFixtu
 be registered with this tag.  
 Please note the Bundle uses `a compiler pass to detect the services
 <https://github.com/doctrine/DoctrineFixturesBundle/blob/HEAD/src/DependencyInjection/CompilerPass/FixturesCompilerPass.php>`_, and in so, 
-they should be instanciable (declaring their constructor as `private` would prevent 
+they should be instanciable (declaring their constructor as ``private`` would prevent 
 them from being detected, for instance).
 
 To see other options for the command, run:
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -73,8 +73,12 @@ Once your fixtures have been written, load them by executing this command:
 
 This command looks for all services tagged with ``doctrine.fixture.orm``. If you're
 using the `default service configuration`_, any class that implements ``ORMFixtureInterface``
-(for example, those extending from ``Fixture``) will automatically be registered
-with this tag.
+(for example, those extending from ``Doctrine\Bundle\FixturesBundle\Fixture``) will automatically 
+be registered with this tag.  
+Please note the Bundle uses `a compiler pass to detect the services
+<https://github.com/doctrine/DoctrineFixturesBundle/blob/HEAD/src/DependencyInjection/CompilerPass/FixturesCompilerPass.php>`_, and in so, 
+they should be instanciable (declaring their constructor as `private` would prevent
```
greg0ire:
```suggestion
they should be instanciable (declaring their constructor as ``private`` would prevent 
```

greg0ire:
:thinking: I don't think we should mention this. I mean, there are compiler passes with such detection in many, many bundles, including official Symfony bundles, right? I don't think they feel compelled to mention this every time.

DocFX:
You mean the fact it's a Cpass? Or the whole thing? Cause this had me blocked for litteral hours until I realized it was the problem. Though it could help.

SenseException:
The compiler pass is an implementation detail and shouldn't matter. The important part is `Doctrine\Bundle\FixturesBundle\Fixture` and autoregister. As far as I know does this imply that a constructor has to be public (autowire), but that's just a guess with my outdated Symfony knowledge.

greg0ire:
> You mean the fact it's a Cpass

No, I mean they don't feel compelled to mention your service has to be instanciable every time they use a compiler pass. IMO this is a problem that isn't specific to this bundle, and should be documented in a more centralized place.

greg0ire:
No answer.

PR #454, opened 2024-11-06
Pass a PSR logger to LoadDataFixturesDoctrineCommand

greg0ire:
Not doing so is deprecated.
See https://github.com/doctrine/data-fixtures/pull/462

### DIFF
--- a/src/Command/LoadDataFixturesDoctrineCommand.php
+++ b/src/Command/LoadDataFixturesDoctrineCommand.php
@@ -12,6 +12,7 @@ use Doctrine\Bundle\FixturesBundle\Purger\PurgerFactory;
 use Doctrine\Common\DataFixtures\Executor\ORMExecutor;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\Persistence\ManagerRegistry;
+use Psr\Log\AbstractLogger;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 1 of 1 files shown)


config/services.xml
```diff
@@ -8,6 +8,8 @@
         <service id="doctrine.fixtures_load_command" class="Doctrine\Bundle\FixturesBundle\Command\LoadDataFixturesDoctrineCommand">
             <argument type="service" id="doctrine.fixtures.loader" />
             <argument type="service" id="doctrine" />
+            <argument type="collection"></argument>
+            <argument type="service" id="logger" on-invalid="null"/>
```
greg0ire: I don't know if this is the right logger instance to use here, @derrabus maybe you will know.

src/Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -34,9 +35,15 @@ class LoadDataFixturesDoctrineCommand extends DoctrineCommand
...
+    private ?LoggerInterface $logger;
+
     /** @param PurgerFactory[] $purgerFactories */
-    public function __construct(SymfonyFixturesLoader $fixturesLoader, ?ManagerRegistry $doctrine = null, array $purgerFactories = [])
-    {
+    public function __construct(
+        SymfonyFixturesLoader $fixturesLoader,
+        ?ManagerRegistry $doctrine = null,
+        array $purgerFactories = [],
+        ?LoggerInterface $logger = null
```
greg0ire: Will deprecate not passing this in the next minor.

src/Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -134,7 +142,7 @@ private function doExecute(InputInterface $input, OutputInterface $output): int
             $input->getOption('purge-with-truncate'),
         );
         $executor = new ORMExecutor($em, $purger);
-        $executor->setLogger(static function ($message) use ($ui): void {
+        $executor->setLogger($this->logger ?? static function ($message) use ($ui): void {
             $ui->text(sprintf('  <comment>></comment> <info>%s</info>', $message));
         });
```
derrabus: Passing a closure as logger is deprecated. We could use an anonymous class instead that extends PSR-3's `AbstractLogger`.

config/services.xml
derrabus: This is fine, I guess. In case, the app uses Monolog, we should subscribe to the Doctrine channel.

```xml
<tag name="monolog.logger" channel="doctrine" />
```

greg0ire: Ah thanks, that was the missing piece 👍

src/Command/LoadDataFixturesDoctrineCommand.php
greg0ire: I don't understand your comment, but I see I made a big mistake here: I didn't check for support of the PSR-3 logger.

greg0ire: I don't see a straightforward way to detect it. I think I'll just bump the dependency.

greg0ire: Ah I understood, Copilot gave me an example implementation. So… I guess I should remove the new `$logger` property entirely then.

composer.json
```diff
@@ -21,7 +21,7 @@
     ],
     "require": {
         "php": "^7.4 || ^8.0",
-        "doctrine/data-fixtures": "^1.3",
+        "doctrine/data-fixtures": "^1.8",
```
greg0ire: If I use `__invoke`, this could be postponed to 3.7 :thinking:

greg0ire:
ping @derrabus @SenseException :pray:

SenseException (approved):

PR #467, opened 2024-11-28
Bump PHPStan to level 7

greg0ire:
I addressed what I could, and the issues I baselined are caused by compatibility layers that should disappear in 4.0.


SenseException (approved):
Not much has changed despite going from level 3 to 7 :+1:

PR #473, opened 2024-11-29
Remove baseline

greg0ire:
This implies bumping `psr/log` to v2, which adds type declarations to its
interface.


greg0ire:
Requires https://github.com/doctrine/DoctrineFixturesBundle/pull/475 , a merge up and a bump of the `psr/log` dependency.

tests/Fixtures/FooBundle/DataFixtures/RequiredConstructorArgsFixtures.php
```diff
@@ -9,6 +9,7 @@
 
 class RequiredConstructorArgsFixtures implements ORMFixtureInterface
 {
+    /** @phpstan-ignore constructor.unusedParameter */
```
greg0ire: This is the only PHPStan issue left in the project, I propose we switch to `phpstan-ignore` for it, and now that PHPStan has error identifiers, maybe we should do that for everything.

SenseException: What's the reason for the constructor to have an argument for the test?

greg0ire: That class was introduced in ee5eb61c41572963299c41489b9acc159cbc246d, which has a commit message that explains the reason.

```
    Also fixing a bug where getDependencies()
    
    Basically, currently getDependencies() won't work if any of your classes
    have required constructor args. We've added a clear error message for
    this situation.
```

SenseException: It's just for a test but maybe the ignore isn't needed when the property is set to public.

SenseException (approved):

PR #480, opened 2024-11-30
Make final classes that can be

greg0ire:
This allows us to adapt to upstream changes without that being a breaking change.


SenseException (approved):

stof:
@greg0ire I also think that some classes (the DI extension, the compiler passes, maybe more) could be tagged as `@internal`, as they are not meant to be used directly by project code.

PR #457, opened 2024-11-13
Allow data-fixtures v2

greg0ire:
The interfaces introduced in 1.5.0 allow to test without mocking a final class.

Closes #456

PR #486, opened 2024-12-11
Fix compatibility for Symfony 6 projects without doctrine/persistence 3 support

alexander-schranz:
This issue libraries run into which require / support still `doctrine/persistence: "^2"` on Symfony 6, see also https://github.com/doctrine/DoctrineBundle/pull/1841 which is the same problem. Symfony 6.4 does not support persistence 3 so 6.3 need still be added to support this correctly to avoid accidently downgrade of the doctrine/bridge which runs into following error.

Before:

> `Compile Error: Declaration of Symfony\Bridge\Doctrine\PropertyInfo\DoctrineExtractor::getProperties(string $class, array $context = []) must be compatible with Symfony\Component\PropertyInfo\PropertyListExtractorInterface::getProperties(string $class, array $context = []): ?arra`

Bildschirmfoto 2024-12-11 um 12 24 59

After:

Bildschirmfoto 2024-12-11 um 12 20 35

Runs correctly.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
         "symfony/console": "^5.4 || ^6.0 || ^7.0",
         "symfony/dependency-injection": "^5.4 || ^6.0 || ^7.0",
         "symfony/deprecation-contracts": "^2.1 || ^3",
-        "symfony/doctrine-bridge": "^5.4.48 || ^6.4.16 || ^7.1.9",
+        "symfony/doctrine-bridge": "^5.4.48 || ~6.3.12 || ^6.4.16 || ^7.1.9",
         "symfony/http-kernel": "^5.4 || ^6.0 || ^7.0"
     },
     "require-dev": {


greg0ire:
The dependency was bumped in https://github.com/doctrine/DoctrineFixturesBundle/pull/457 to `^6.4.16` in order to obtain https://github.com/symfony/symfony/pull/58865

So although it does not show in the build because we are not testing all possibilities, I think with this PR, you will be breaking support for data-fixtures 2.

Also, 6.3 is no longer supported.

And finally, this looks like an error that should be fixed at the Symfony level.

alexander-schranz:
> So although it does not show in the build because we are not testing all possibilities, I think with this PR, you will be breaking support for data-fixtures 2.

The issue is only when using doctrine/persistence 2. You will not get data-fixtures 2 on doctrine/persistence 2 so as we still allow 5.4 we should also be able to allow 6.3. I did a quicktest on https://github.com/sulu/sulu/pull/7699 and didn't run into any issues there.

> Also, 6.3 is no longer supported.

I know but required if you can not yet update to doctrine/persistence 3.0 yet because of bc reasons.

greg0ire:
> as we still allow 5.4 we should also be able to allow 6.3

Why? Sorry but I don't see the logical connection.

alexander-schranz:
> Why? Sorry but I don't see the logical connection.

Did just see now that the fix also part of 5.4 but not 6.3 thought it was 6.4 and above as it was for data-fixtures 2.

But I think it is still not a problem as on doctrine persistence 2 on Symfony 6.3 it uses data-fixtures 1 is there https://github.com/symfony/symfony/pull/58865 is not required and as you see in https://github.com/sulu/sulu/pull/7699 it works like expected.

Without allowing 6.3 persistence doctrine-bridge will get downgraded and fail with: https://github.com/sulu/sulu/actions/runs/12050408254/job/34248387183

greg0ire:
On the failing build you linked, I can see that `doctrine-bridge` 5.4 is being installed alongside with `symfony/property-info` 7.2.

If the cause of the issue is that `doctrine-bridge` 5.4 is not compatible with `symfony/property-info` 7.2, maybe that is what should be fixed.

And if yes it should be fixed a bit like I did in https://github.com/symfony/symfony/pull/58865, with a dynamic trait.

greg0ire:
I think a PR equivalent to https://github.com/symfony/symfony/pull/41291, but for Symfony 7 should have been done but wasn't, and now 5.4 is in security fixes mode only: https://symfony.com/releases

So… I think it will not be accepted, since it has nothing to do with security.

This is a Symfony bug though, right? So… people using 5.4 and encountering this bug should upgrade.

greg0ire:
Since 5.4 is in security mode, maybe we could simply drop support for 5.4 in this bundle? That way, it will not get downgraded.

derrabus:
This issue will be solved when symfony/symfony#59185 gets released, so let's close this PR.

alexander-schranz:
Can confirm with the latest Symfony release we are able to use Symfony 6.4 with doctrine/persistence 2 and so a few projects can now easily upgrade to 6.4 and not longer run into this kind of issues. Thank you all!

PR #497, opened 2025-03-22
Setup dependabot for PHP

greg0ire:
Pinning our development tools means we should regularly bump their version. Let us automate that.


greg0ire:
The build should get fixed when https://github.com/doctrine/DoctrineFixturesBundle/pull/496 is merged up.

SenseException (approved):

PR #498, opened 2025-03-24
tests: add cache-dir to phpunit.xml.dist

Chris53897:
https://github.com/doctrine/DoctrineFixturesBundle/actions/runs/14010902231/job/39230551312?pr=495


greg0ire:
I think this is a bug in PHPUnit, and  we should rather pin the dependencies, like so: https://github.com/doctrine/DoctrineFixturesBundle/pull/496

greg0ire:
Fixed by https://github.com/doctrine/DoctrineFixturesBundle/pull/499, thanks!

PR #499, opened 2025-03-24
Merge 3.7.x up into 4.0.x

greg0ire:


composer.json
```diff
@@ -37,9 +37,9 @@
         "symfony/http-kernel": "^5.4 || ^6.0 || ^7.0"
     },
     "require-dev": {
-        "doctrine/coding-standard": "^12",
-        "phpstan/phpstan": "^2",
-        "phpunit/phpunit": "^10.5.38 || ^11"
+        "doctrine/coding-standard": "12.0.0",
+        "phpstan/phpstan": "2.1.8",
+        "phpunit/phpunit": "^10.5.38 || 11.4.14"
```
alexisLefebvre, line 42:
```suggestion
        "phpunit/phpunit": "^10.5.38 || ^11.4.14"
```

Was it necessary to not use a version range?

greg0ire: Yes, see af9f8c7

PR #510, opened 2025-04-15
Allow fixtures loader custom implementation

kshtompel:
- Allow fixtures loader custom implementation


greg0ire:

kshtompel:
@greg0ire please suggest me what to do here, because I see no problem in code.
Interface is declared and imported.
image

src/Loader/FixturesProviderInterface
greg0ire, line 1: Oh I don't know... Have you tried putting this in a file that ends with `.php`? Might work slightly better...

kshtompel: Oh, thank you. didn't get it

greg0ire:
Let me insist on this:

> It seems there are CI jobs failing. Please take a look at [this guide](https://www.doctrine-project.org/contribute/#dealing-with-checks-and-tools) for more on how to handle those.

kshtompel:
@greg0ire thanks for the guidance. All checks are passed

config/services.xml
```diff
@@ -6,14 +6,13 @@
...
+            <argument type="service" id="doctrine.fixtures.provider" />
             <argument type="service" id="doctrine" />
             <tag name="console.command" command="doctrine:fixtures:load" />
         </service>
 
-        <service id="doctrine.fixtures.loader" class="Doctrine\Bundle\FixturesBundle\Loader\SymfonyFixturesLoader" public="false">
-            <argument type="service" id="service_container" />
-        </service>
+        <service id="doctrine.fixtures.provider" alias="doctrine.fixtures.loader" />
+        <service id="doctrine.fixtures.loader" class="Doctrine\Bundle\FixturesBundle\Loader\SymfonyFixturesLoader" public="false" />
```
greg0ire: (why) is this change necessary? Or maybe is it just a best practice?

kshtompel: As LoadDataFixturesDoctrineCommand class right now receives FixturesProvider interface not SymfonyFixturesLoader we should define it as separate service to allow it to be overridden and as it the same service as SymfonyFixturesLoader I just put alias on it for simplicity reason only

greg0ire: OK, but isn't it possible for you to override `doctrine.fixtures.loader` just by defining a service with that `id` in your application? I think I remember doing something like that in the past.

kshtompel: it can be overridden only if it was defined otherwise it should be a required part of bundle configuration to be set up

greg0ire: According to what you just wrote, since `doctrine.fixtures.loader` is defined, it can be overridden, so this is useless, but I feel this isn't what you mean to say.

kshtompel: SymfonyFixturesLoader class is wider implementation then FixturesProvider interface and it is used to configure bundle, see [FixturesCompilerPass](https://github.com/doctrine/DoctrineFixturesBundle/blob/4.1.x/src/DependencyInjection/CompilerPass/FixturesCompilerPass.php) and [\Doctrine\Common\DataFixtures\Loader](https://github.com/doctrine/data-fixtures/blob/2.0.x/src/Loader.php). So If you want to override doctrine.fixtures.loader you need implement additional methods in your implementation which are not defined in new interface. So as I mentioned I defined new interface and new service which is just alias for current fixtures loader implementation for simplicity.

greg0ire: >  and it is used to configure bundle, see FixturesCompilerPass and \Doctrine\Common\DataFixtures\Loader.

I'm AFK until Monday, and can only access this on my phone. Please provide GitHub permalinks

kshtompel: updated previous comment with links to repos

greg0ire (approved):

config/services.xml
SenseException: Because you changed this line, you can remove `public="false"` because afaik are Symfony services private by default.

SenseException (changes requested):

SenseException (approved):

greg0ire:
Thanks @kshtompel

PR #519, opened 2025-10-03
Migrate service config from XML to PHP

mbabker:
Addresses deprecations from https://github.com/symfony/symfony/pull/60568


stof (approved):

stof:
@mbabker are you doing such conversions by hand or are you using the script from https://github.com/GromNaN/symfony-config-xml-to-php/ ?

mbabker:
> @mbabker are you doing such conversions by hand or are you using the script from https://github.com/GromNaN/symfony-config-xml-to-php/ ?

The PRs I've done have been by hand since I'm just scanning the app I'm working in for simpler bundle configs right now (could probably try it with the migrations bundle and see how it goes).

PR #522, opened 2025-10-12
Add support for `doctrine/doctrine-bundle` 3.0

dmaicher:


dmaicher:
Or should I target 4.2.x?

Edit: done. Makes more sense,.

composer.json
```diff
@@ -39,7 +39,7 @@
     "require-dev": {
         "doctrine/coding-standard": "13.0.0",
         "phpstan/phpstan": "2.1.11",
-        "phpunit/phpunit": "^10.5.38 || 11.4.14"
+        "phpunit/phpunit": "^10.5.38 || ^11.4.14"
```
greg0ire: This change should happen on lower branches as well, if it's really what is meant. But PHPStan does not have `^` either, so it's probably meant to be bumped by dependabot… which might not work 😬 
Anyway, there is no reason to have it in this PR.

dmaicher: reverted :+1:

greg0ire (approved):

greg0ire:
Thanks @dmaicher !

PR #541, opened 2025-11-04
allow DoctrineBundle 4.0

xabbuh:
this is required for the bundle to be usable with Symfony 8.0


greg0ire:
Looks like 3.1 would be enough: https://github.com/doctrine/DoctrineBundle/pull/2069, and it's already covered by the current version constraint, so I think we should close this.

xabbuh:
Oh I missed that, so looks like #542 doesn't need these changes indeed.

PR #548, opened 2025-11-19
test: allow PHPUnit 12

Chris53897:
I did not add `^` like in the version 11

If i see this correct the CI before always uses `10.5.58` and not `11.4.14`.


greg0ire:
Should target 3.7

Chris53897:
https://github.com/doctrine/DoctrineFixturesBundle/pull/550

PR #549, opened 2025-11-19
ci: add testrun for PHP 8.5

Chris53897:


greg0ire:
Oh this should have targeted 3.7

PR #565, opened 2025-12-03
Flag 3.7.x as unmaintained

derrabus:
We've released 4.0.0 a year ago and I believe the migration path from 3.7 is easy enough. Let's stop maintaining two major branches of this bundle in order to keep things simple for us.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,11 +8,11 @@ updates:
     open-pull-requests-limit: 10
     labels:
       - "Dependencies"
-    target-branch: "3.7.x"
+    target-branch: "4.3.x"
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
       interval: "weekly"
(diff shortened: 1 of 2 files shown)


stof (approved):

greg0ire:
Are you sure about this? There are more downloads of v3 than there are downloads of v4

Screenshot 2025-12-03 at 22-14-57 Install Statistics - doctrine_doctrine-fixtures-bundle - Packagist

derrabus:
Yes, and many of them are still on 3.4. 🤷‍♂️  My impression is that people just didn't upgrade the bundle because they didn't have to.

SenseException (approved):

greg0ire:
OK, maybe this will contribute to nudging them in the right direction.

greg0ire:
Now that this is merged, I think we should remove the 3.x branches.

